Issue #33, opened 2016-01-09
unable to access options on hoverzoom+


frogbong1:
it wont let me click on any options

extesy:
I can't reproduce that so I need more details from you. What is your OS and browser version? Can you try opening Chrome dev tools and see if there are any errors on the console tab?

AltarBeastiful:
I get an error from gumby when opening the options :
```javascript
Uncaught ReferenceError: yepnope is not defined
gumby.min.js:1
```
Tested on Windows 10, chrome version 47.0.2526.106 m, HoverZoom+ 1.0.43

extesy:
This is weird because options worked for me and majority of other users. I've merged the pull request anyways, hopefully that would help. Please try version 1.0.44.

frogbong1:
Im using windows 10 and version 1.0.44 and chrome is up to date. when I open the options page it shows "general, sites, action keys and advanced" I can't click on any of it

frogbong1:
I just restarted my LT and now its working.

PR #35, opened 2016-01-10
Revert "Update dependency versions"

AltarBeastiful:
Revert a commit updating depedencies and causing #33.

Modernizr latest versions deprecates some yepnodes related functions needed in Gumpy. As Gumpy went unmaintened it's unlikely to see update on their part. We should avoid to update those depedencies if possible.
For more information see : https://github.com/SlexAxton/yepnope.js#deprecation-notice

Now options works but I get other errors when opening options :
```
js/libs/jquery.mobile.custom.min.js:1 GET chrome-extension://oahkfgcalecafmamfcaombebonibgfmd/html/js/libs/jquery.mobile.custom.min.js net::ERR_FILE_NOT_FOUND
modernizr-2.7.1.min.js:1 GET chrome-extension://oahkfgcalecafmamfcaombebonibgfmd/html/js/libs/jquery.mobile.custom.min.js net::ERR_FILE_NOT_FOUNDa.yepnope.injectJs @ modernizr-2.7.1.min.js:1(anonymous function) @ modernizr-2.7.1.min.js:1
```

Apparently JQuery mobile and Modernizr are trying to find `jquery.mobile.custom.min.js` in `html/js/libs/` instead of just `js/libs/`.
Is that what you were trying to fix?


extesy:
Thanks for identifying the root cause. Once bootstrap v4 is out, I'll replace gumby with it, so modernizr wouldn't be required.

PR #101, opened 2016-06-18
Fixed Null and Null; bug in Imgur

itpun:
Hi! 

I noticed that there was a bug in Hoverzoom where if there was an album with a mix of  captions and no captions, there are cases where it would spill "null" and "null;" characters. Im thinking this may have been caused by some parsing issue with newlines and dealing with no characters, but I did a quick fix just to clean the captions of nulls before showing the frame. I tried to use regex instead of "null" but it didn't seem to work. Feel free to lemme know if you have any ideas to make this a little cleaner.

Example of bug:
1) https://www.reddit.com/r/gaming/comments/4ohlaj/zelda_breath_of_the_wild_actually_has_a_fully/
2) hover over the thumbnail
3) scroll a few pages through, you should see the nulls.


plugins/imgur_a.js
```diff
@@ -76,7 +84,7 @@ hoverZoomPlugins.push({
                                                 caption += ';\n';
                                             }
                                             caption += img.description;
```
extesy: I think nulls are coming from this line. If it's possible to filter them here (just compare img.description with "null"), then you don't need to do replace later.

Issue #72, opened 2016-04-23
Feature Request: Color picker for frame


STaRDoGG:
Small req to add an option to select the color of the frame around the popup that shows the image/video. Currently it's white; I'd love to be able to tweak it to match my browser theme(s).

stuntguy3000:
Agreed, would be good for things like "night modes"

itpun:
I currently have a pull request that will allow you to have a Dark Mode setting. It is totally possible to do custom border colors, but I would like your opinion on having set color themes or to pick from a RBG color wheel.

extesy:
@itpun I prefer generic (not one-off) solutions, and ability to choose custom color (with default value #fff) satisfies both this request and a request for a dark theme. But don't forget that the text color should be chosen automatically to appear good on any random backgound color.

extesy:
Implemented in 1.0.124

PR #103, opened 2016-06-19
Dark mode feature

itpun:
Dark mode is now a feature for HoverZoom. 

Option to set Dark mode is currently in "Advanced" settings. When enabled, the image border color will be set to black and the caption color will be set to white. I can further stylize it later if needed.

I apologize I think i screwd up my branching, this branch might also include my imgur null fix. Please let me know what I can do to fix this if you want!

Ian


JamieSharpe:
Is it feasible to allow a custom colour to be set?

Then it would also bring in custom themes in general, and not just the hard coded colours.

See issue/request https://github.com/extesy/hoverzoom/issues/72.

itpun:
Hey it's totally possible. Currently I'm wondering if its best to have preset standard of some themes first, or let the user pick whatever colors they want.

JamieSharpe:
Hi,

I think that would be the best approach. Having a small selection of preset colours along with textbox/range sliders for RGB custom colours too.

js/hoverzoom.js
```diff
@@ -449,13 +452,41 @@ var hoverZoom = {
                     video.src = imgDetails.url;
                     imgFullSize = $(video).appendTo(hz.hzImg);
                     video.addEventListener('error', imgFullSizeOnError);
-                    video.addEventListener('loadedmetadata', posImg);
+                    video.addEventListener('loadedmetadata', function(){
+                        posImg();
+                        track = this.addTextTrack("captions", "English", "en");
```
extesy: Extract this into a function and put behind a new option (disabled by default). Changes like this shouldn't suddenly appear for 280k current users.

js/hoverzoom.js
```diff
@@ -449,13 +452,41 @@ var hoverZoom = {
...
+                                track.addCue(new VTTCue(time, time+1, finalTime)); 
+                                track.cues[time].align ="end";
+                                track.cues[time].position = 100;
+                                track.cues[time].line = 0;
+                            }
+                        }
+                    });
                     video.addEventListener('loadeddata', function() {
-                        imgFullSizeOnLoad();
+                        ();
```
extesy: what does this mean?

js/hoverzoom.js
```diff
@@ -140,6 +140,9 @@ var hoverZoom = {
                     if (hzCaption.height() > 20) {
                         hzCaption.css('font-weight', 'normal');
                     }
+                    if(options.enableDarkMode) {
```
extesy: please respect existing code style - have a single space on both sides of parentheses.

js/hoverzoom.js
```diff
@@ -89,7 +89,7 @@ var hoverZoom = {
                 'font':'menu',
                 'font-size':'11px',
                 'font-weight':'bold',
-                'color':'#333',
+                'color':'#000000',
```
extesy: why change this? pure black color makes contrast too high.

plugins/imgur_a.js
```diff
@@ -76,7 +84,7 @@ hoverZoomPlugins.push({
                                                 caption += ';\n';
                                             }
                                             caption += img.description;
```
extesy: I think that "null"s are appended on this line so it's better to check here rather than do a replace later.

extesy:
I've cherry-picked a code to show video timestamps, but the dark feature should be reworked into a more generic color picker, so I'm closing this PR.

PR #104, opened 2016-06-19
Feature option video range

JamieSharpe:
Added a range slider to the video option volume.


JamieSharpe:
Video range slider in options page

css/style.css
```diff
@@ -3,3 +3,26 @@
     margin-top: 2em;
     margin-bottom: 2em;
 }
+
+.append input[type=range] + input[type=text] {
+    border-radius: 0px;
+    -webkit-border-radius: 0px;
```
extesy: why do you need -webkit attributes? this extension is only supported on latest versions of Chrome, so border-radius alone should be fine.

html/options.html
```diff
@@ -110,6 +110,7 @@ <h1 class="ten columns text-center" data-i18n="optTitle"></h1>
                             <div class="ttip" data-i18n-tooltip="optVideoVolumeTooltip">
                                 <li class="append field"><label class="inline" for="txtVideoVolume">
                                     <div style="display:inline" data-i18n="optVideoVolume"></div>
+                                    <input class="xnarrow input" type="range" name="points" min="0" max="100" step="1" id="rngVideoVolume" >
```
extesy: please remove a space before closing >

extesy:
Thanks for the PR, @JamieSharpe!

JamieSharpe:
Changed and tested with removed webkit attribute, also addressed the syntax styles.

PR #117, opened 2016-08-05
Initial update for edge

rakion99:
sites tested: imgur, giphy, catgifpage, i.redd.it. gfycat dont work
with only these changes you can load this extension to Microsoft Edge and will work but not all, like the options page and other things like extension not loading on MS Edge restart but its a progress :3.
in the manifest.json i changed page_action to browser_action so the icon is in the side of the url bar in MS Edge
Maybe we need other branch for Microsoft Edge extension???
#114


rakion99:
some documentation from Microsoft https://developer.microsoft.com/en-us/microsoft-edge/platform/documentation/extensions/

PR #168, opened 2016-11-10
Updated FetLife with Cache Tokens

iambic9:
FetLife updated their image serving to require some kind of CSRF-style token to view any image.  This loads the actual picture page in order to get the full source (with token).  Previously just swapping the resolution suffix was enough.

---
This change is [Reviewable](https://reviewable.io/reviews/extesy/hoverzoom/168)


extesy:
Thanks for the contribution, @iambic9!

PR #172, opened 2016-11-14
Added support for clips.twitch.tv

Markussss:
While adding support for clips.twitch.tv, I looked at the method used
in both plugins/reddit.js and plugins/gfycat.js. I think it turned out
right, but I'm not 100% sure I did it "by the books". But it works, so
I've that going at least.

---
This change is [Reviewable](https://reviewable.io/reviews/extesy/hoverzoom/172)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -890,7 +890,8 @@
                 "plugins/memefive.js",
                 "plugins/gfycat.js",
                 "plugins/pornbot.js",
-                "plugins/memedad.js"
+                "plugins/memedad.js",
+                "plugins/clips.twitch.js"
             ]
         },
         {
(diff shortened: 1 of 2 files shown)


extesy:
I wonder if it is possible to compute full video url without having to load twitch page, just from using a source preview url. @Markussss, can you please provide an example url on reddit that links to clips.twitch.tv?

Markussss:
@extesy Sure! There are a lot on /r/Dota2, if you want more tests, but here is an example url: https://clips.twitch.tv/admiralbulldog/BrainySalamanderShazBotstix.

I was hesitant about loading the twitch page to get the full video url, but after looking through the twitch API docs and playing around with some code, I couldn't see any other ways to do it.

Maybe it could be possible to reverse the four-word ID in some way?

PR #179, opened 2016-11-18
Fix of plugin: 1x

GrosPoulet:
Hi

This is my first pull request for HoverZoom, and hopefully not the last :fearful:
I fixed the js plugin: 1x.js
Fix was tested on 1x.com, it should have no impact anywhere else since plugin is only loaded on "*://*.1x.com/*" urls (according to manifest).

I'm new to GitHub (and to js!), so:
- please double-check!  
- comments/advices warmly welcomed!

Have a nice day :smiley: 

Fix details:
Added new search patterns:
-ld.
-sd.
-hd2.
-sq.
-square.

---
This change is [Reviewable](https://reviewable.io/reviews/extesy/hoverzoom/179)

### DIFF
--- a/plugins/1x.js
+++ b/plugins/1x.js
@@ -1,7 +1,7 @@
 var hoverZoomPlugins = hoverZoomPlugins || [];
 hoverZoomPlugins.push({
     name:'1x.com',
-    version:'0.1',
+    version:'0.2',
     prepareImgLinks:function (callback) {
         var res = [];
         hoverZoom.urlReplace(res,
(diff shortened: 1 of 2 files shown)


GrosPoulet:
Fix of plugin: 1x

plugins/1x.js
```diff
@@ -24,6 +24,31 @@ hoverZoomPlugins.push({
             '-thumb.',
             '-fullsize.'
         );
+		hoverZoom.urlReplace(res,
```
extesy: Please use spaces instead of tabs in the source code. Take a look how this diff looks in github (https://github.com/extesy/hoverzoom/pull/179/files) and you'll easily see the difference.

plugins/1x.js
```diff
@@ -24,6 +24,31 @@ hoverZoomPlugins.push({
...
+            'img[src]',
+            '-ld.',
+            '-hd2.'
+        );
+		hoverZoom.urlReplace(res,
```
extesy: This replace is not necessary. When the previous one (on line 27) completes, you'll no longer have any images with `-ld.` in them, so this replacement won't do anything.

plugins/1x.js
```diff
@@ -24,6 +24,31 @@ hoverZoomPlugins.push({
...
+            'img[src]',
+            '-ld.',
+            '-sd.'
+        );
+		hoverZoom.urlReplace(res,
+            'img[src]',
+            '-sq.',
+            '-hd2.'
+        );
+		hoverZoom.urlReplace(res,
```
extesy: This replace is not necessary. When the previous one (on line 37) completes, you'll no longer have any images with `-sq.` in them, so this replacement won't do anything.

GrosPoulet:
Hi
I replaced tabulations by spaces in **1x.js plugin**.
Did you have time to check the others issues?

PR #181, opened 2016-11-22
Now loads youtube video with the highest resolution available

grubeb2:
Zoomed images and videos now fill the window when the full zoom key is pressed.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -173,7 +173,7 @@ var hoverZoom = {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


extesy:
I don't think it's a good idea. Some youtube videos have resolutions up to 4K and using those for **previews** is overkill. I used 720p for previews specifically as a compromise between quality and traffic consumption. How about checking `options.showHighRes` option instead like it's done in other plugins?

extesy:
Also could you please squash your commits into one?

PR #187, opened 2016-12-02
Master

GrosPoulet:
Fix for plugin: deviantart.js


GrosPoulet:
Fix for plugin: deviantart.js

PR #197, opened 2016-12-23
Master

GrosPoulet:
Added plugin for Fotoblur.com


GrosPoulet:
OK, should work on all images now

PR #227, opened 2017-02-05
Added plugin for: Foursquare.com

GrosPoulet:
Find the best places to eat, drink, shop, or visit in any city in the world. Access over 75 million short tips from local experts.


plugins/foursquare.js
```diff
@@ -0,0 +1,22 @@
+var hoverZoomPlugins = hoverZoomPlugins || [];
+hoverZoomPlugins.push({
+    name:'cgcookie.com',
```
extesy: @GrosPoulet Can you please fix the plugin name?

PR #233, opened 2017-02-15
Added plugin for: Wookmark.com

GrosPoulet:
Wookmark is an online tool for collecting images.


extesy:
@GrosPoulet Looks like metmuseum.js plugin is already added, so it shouldn't be in this pull request. Can you please sync with master and resolve conflicts?

PR #250, opened 2017-03-20
Prevent hoverzooming from removing titles of all titled things.

Eegras:
There was a typo.  `this` referred to all titledElements.  `$(this)` refers to only the hover-zoomed thing.


extesy:
Thanks for the fix!

PR #274, opened 2017-07-30
Firefox support

Mte90:
With this fix the extension works without problems on Firefox

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -106,10 +106,13 @@ function showPageAction(tab) {
 // Checks if the extension has been updated.
 // Displays a notification if necessary.
 function checkUpdate() {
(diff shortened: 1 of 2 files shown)


extesy:
@Mte90 why change the extension package name?

Mte90:
If you do now `npm install` you get an error for a wrong name that contain symbols :-)

extesy:
Why do you need to run `npm install`? Changing package name might have unintended consequences for Chrome Webstore and I would prefer to avoid that risk.

Mte90:
I changed the name on package.json that is used by `npm` the chrome webstore use manifest.json :-)

extesy:
I see, my bad. Looks good to me!

PR #277, opened 2017-07-30
Update jquery

Mte90:
Actually I updated to jquery 2.2.4 because to 3.2.1 I got few errors.


manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "__MSG_extName__",
+    "name": "HoverZoom+",
```
extesy: this is required for localization

Mte90: You are right, I fixed :-)

Mte90:
Finally finished the porting to jquery 3.2.1

Mte90:
Seems that on google plugin there are issue but I am working on it

Mte90:
I tested also on chrome and works without problems now (the google plugin now use the right selector)

Issue #300, opened 2017-09-22
Warnings raised by addons.mozilla.org while publishing


stil:
1. **"tabs.getAllInWindow" is deprecated or unimplemented**

    **Warning:** This API has been deprecated by Chrome and has not been implemented by Firefox.
    ```
    js/common.js line 64 column 13
    ```
    

1. **Unsafe assignment to innerHTML**

    **Warning:** Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
    ```
    js/libs/modernizr-2.7.1.min.js line 1 column 2026
    ```
    

1. **Unsafe assignment to innerHTML**

    **Warning:** Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
    ```
    js/libs/modernizr-2.7.1.min.js line 1 column 8197
    ```
    

1. **The Function constructor is eval.**

    **Warning:** Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all possible.'
    ```
    js/libs/modernizr-2.7.1.min.js line 1 column 9049
    ```
    

1. **"app.getDetails" is deprecated or unimplemented**

    **Warning:** This API has been deprecated by Chrome and has not been implemented by Firefox.
    ```
    js/options.js line 293 column 76
    ```
    

1. **eval can be harmful.**

    **Warning:** Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all possible.'
    ```
    plugins/sky_it.js line 19 column 21
    ```
    

1. **"tabs.getSelected" is deprecated or unimplemented**

    **Warning:** This API has been deprecated by Chrome and has not been implemented by Firefox.
    ```
    js/background.js line 78 column 13
    ```
    

1. **"app.getDetails" is deprecated or unimplemented**

    **Warning:** This API has been deprecated by Chrome and has not been implemented by Firefox.
    ```
    js/background.js line 111 column 27
    ```
    

1. **"tabs.getSelected" is deprecated or unimplemented**

    **Warning:** This API has been deprecated by Chrome and has not been implemented by Firefox.
    ```
    js/popup.js line 47 column 5
    ```
    

1. **Unsafe assignment to innerHTML**

[...]

PR #301, opened 2017-09-22
Fix some of addons.mozilla.org warnings

stil:
This PR fixes following warnings from #300 

1. "tabs.getAllInWindow" is deprecated or unimplemented
5. "app.getDetails" is deprecated or unimplemented
7. "tabs.getSelected" is deprecated or unimplemented
8. "app.getDetails" is deprecated or unimplemented
9. tabs.getSelected" is deprecated or unimplemented

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -75,7 +75,7 @@ function onMessage(message, sender, callback) {
             });
             break;
         case 'openViewTab':
-            chrome.tabs.getSelected(null, function (currentTab) {
+            chrome.tabs.query({active: true}, function (currentTab) {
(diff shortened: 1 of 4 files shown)


extesy:
@stil Thank you for the pull request. Have you validated that everything still works in Chrome?

stil:
Yes, I checked briefly on Chrome and it works fine. For example before this fix Firefox wouldn't display this notice on Options page (on very bottom):

````
Version 1.0.124 - © 2014-2016 Oleg Anashkin
```

Right now it works fine in both browsers.

PR #309, opened 2017-10-10
Add clippituser.tv plugin

mrdziuban:
Adds a plugin for https://clippituser.tv/ which I've started to see in some reddit threads as an alternative to streamable.


extesy:
Thanks for the contribution, Matt!

PR #329, opened 2018-01-06
Use unminified version of 3rd party libs to comply with policies

stil:
This PR replaces minified JS libraries with unminified ones.
AMO (addons.mozilla.org) can take down the addon if it includes minified libraries.

I have tested it and it seems to work fine. I'll leave this PR for a day or two to discuss and then I'll merge it.


extesy:
Looks good to me 👍

PR #381, opened 2018-06-12
Instagram: use full-size image instead of the default one.

lapo-luchini:
Use a trick that replaces the post URL with an easy to remember URL that automatically redirects to the full-size image, instead of the (inexplicably) low-rez one that Instagram uses.


extesy:
Thanks for the contribution, @lapo-luchini! I'll test this and merge soon.

lapo-luchini:
I just noticed (in other plugins) the existence of `if (options.showHighRes)`, do you think it makes sense to use it in this case too?

extesy:
@lapo-luchini Yeah, makes sense. Do you want to submit a pull request?

lapo-luchini:
Sure thing.

PR #384, opened 2018-06-19
Enhanced Download, No need for permission and fixed Save Dialog, Fixed Gfycat plugin

ivan2340:


ivan2340:
Also fixed the twitch clips plugin

extesy:
Nice! I'll test this at home and merge later today.

PR #385, opened 2018-06-20
Critical hotfix, hardcoded url

ivan2340:


ivan2340:
Sorry don't know how this snuck in

ivan2340:
Please merge this ASAP before any "damage" occurs

ivan2340:
i know there *might* be some weird url that breaks this, however there's now two ways of parsing for a filename if one of them fails the other takes over, and anyways saving as "image.jpg" is not really a good idea anyways

PR #393, opened 2018-07-06
Fixed loadNextGalleryImage() recursing infinitely on error

KazanK:
In `rotateGalleryImg(rot)`, if the next `hoverZoomGallerySrc` element is an array of possible URLs to try,
`loadNextGalleryImage()` will recurse infinitely on error (i.e. when the URL it's trying to reach is invalid). When attaching handlers to `imgFullSize` in `loadNextGalleryImage()`, the error handler should probably be `imgFullSizeOnError` instead of recursing back into `loadNextGalleryImage`.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1286,7 +1286,7 @@ var hoverZoom = {
         function loadNextGalleryImage() {
             clearTimeout(loadFullSizeImageTimeout);
             imgDetails.url = hz.currentLink.data().hoverZoomSrc[hz.currentLink.data().hoverZoomSrcIndex];
-            imgFullSize.on('load', nextGalleryImageOnLoad).on('error', loadNextGalleryImage).attr('src', imgDetails.url);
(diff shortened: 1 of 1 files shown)


extesy:
@KazanK I'm not sure if this specific fix is a good idea. Infinite recursion is definitely bad but the intention was to try the next image if the current one fails. Most often _some_ image in the array will load successfully. I think what needs to happen here is `onError` handler should increment `hoverZoomSrcIndex` before calling `loadNextGalleryImage` and only if `hoverZoomSrcIndex` is more than array's length then call `imgFullSizeOnError` instead.

KazanK:
Doesn't `imgFullSizeOnError` already do that though? I'm working on a new plugin, and this fix seems to do the trick. I've since changed how the plugin works, so it doesn't have to guess at URLs, but the only issue I had with this fix was that the HoverZoomed image would disappear for a moment before correctly displaying the next image.

extesy:
@KazanK My bad, you are right. Thanks for the pull request! 👍

PR #431, opened 2019-02-15
Handle v.redd.it videos reliably and support audio

mrdziuban:
I've found that a lot of videos on v.redd.it won't load with hoverzoom because the URL `/DASH_600_K` doesn't exist for that video. This updates the code that handles getting video links to request the `DASHPlaylist.mpd` file, which contains info about the available video resolutions, and then choose the highest resolution.

Additionally, the `DASHPlaylist.mpd` will contain info about audio streams, when available, so I've added functionality to support playing audio from a separate source URL than the video. When a separate audio source is specified via the `hoverZoomAudioSrc` data key, an HTML `audio` element will be created along with the `video`, and their playbacks will be kept in sync.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -51,6 +51,7 @@ var hoverZoom = {
 
         var imgDetails = {
                 url:'',
+                audioUrl:'',
                 host:'',
                 naturalHeight:0,
                 naturalWidth:0,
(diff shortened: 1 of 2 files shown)


extesy:
Thank you for the contribution, @mrdziuban! Let me test this first and then I'll merge it shortly.

PR #432, opened 2019-02-26
Fix streamable

mrdziuban:
Streamable videos (e.g. https://streamable.com/rq6ui) no longer have a `source` element, the source of the video is just on the `video` element.


extesy:
@mrdziuban Thank you!

Issue #455, opened 2019-08-17
profile images on new twitter no longer loads


Sushubh:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem
Hoverzoom works great on old Twitter.com. But on new Twitter.com it is now struggling to load profile images on mouse hover. Surprisingly, this was not the case few days ago even on new Twitter.com. Seems like a recent update on Twitter.com broke the app. 

## Url of the page where the problem occurs (not the image url)
Twitter.com

## Steps to reproduce the problem (include image url here)
1. Load Twitter.com
2. Hover mouse cursor over any user profile image
3. Twitter loads the hovercard for that user but hoverzeem does not load preview the profile image. 

## OS version, extension version (Chrome only, Firefox is not supported)
Windows 10, Chrome 77.0.3865.35.

PR #460, opened 2019-09-19
#455: Fix for profile images on Twitter no longer loading

mmakarwich10:
- Added another urlReplace on twimg.js to account for the new mapping for the thumbnail and the full-size image.

- Also turned off the pointer-events for the new div that is covering the profile thumbs, and is preventing hoverzoom from accessing the  element.

### DIFF
--- a/plugins/twimg.js
+++ b/plugins/twimg.js
@@ -14,7 +14,7 @@ hoverZoomPlugins.push({
             /_bigger/,
             '_400x400'
         );
-        $('div.css-1dbjc4n.r-1twgtwe.r-sdzlij.r-rs99b7.r-1p0dtai.r-1mi75qu.r-1d2f490.r-u8s1d.r-zchlnj.r-ipm5af').css('pointer-events', 'none');
+        $('a[aria-haspopup="false"] > div:nth-child(2)').css('pointer-events', 'none');
         callback($(res));
     }
 });
--- a/plugins/twimg.js
+++ b/plugins/twimg.js
@@ -1,7 +1,7 @@
 ﻿var hoverZoomPlugins = hoverZoomPlugins || [];
 hoverZoomPlugins.push({
     name:'Twimg.com',
-    version:'0.1',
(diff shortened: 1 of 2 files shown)


plugins/twimg.js
```diff
@@ -1,14 +1,20 @@
...
             'img[src*="twimg.com"]',
             /_(normal|mini)/,
             ''
         );
+        hoverZoom.urlReplace(res,
+            'img[src*="twimg.com"]',
+            /_bigger/,
+            '_400x400'
+        );
+        $('div.css-1dbjc4n.r-1twgtwe.r-sdzlij.r-rs99b7.r-1p0dtai.r-1mi75qu.r-1d2f490.r-u8s1d.r-zchlnj.r-ipm5af').css('pointer-events', 'none');
```
extesy: I'm not sure how stable this change would be. These class names look auto-generated during compile time so they might completely change every time Twitter updates its website. Is there any other way to find the element?

mmakarwich10: Yeah, I thought so too, but rather I thought it changed on login (it doesn't).

But, you could be right, could change on recompile.

I did try to access it with the "nth-child" pseudo element, but apparently JQuery doesn't support that.

I'll look at it again to see if I can find a better way to access it.

extesy: You don't have to find exact element with a single selector. Fetching multiple candidates and filtering through them using some custom javascript would be fine too.

mmakarwich10: True.

Apparently I'm just crazy, I was wrong about JQuery not supporting the "nth-child" pseudo element. I may have been thinking about another project I'm working on that has an older version of JQuery.

Anyway, fixed the selector. Let me know what you think.

extesy: Looks good to me, thank you for the contribution!

mmakarwich10: No prob! I love this extension and I'm happy to help out!

PR #509, opened 2020-05-11
Fix for 500px plug-in

GrosPoulet:
Fix for 500px plug-in


GrosPoulet:
API calls added in order to obtain fullsize urls.

PR #516, opened 2020-05-20
Fix for ebay plug-in

GrosPoulet:
Fix for ebay plug-in


extesy:
@GrosPoulet Is boardgamegeek change supposed to be here or is it a merge conflict?

GrosPoulet:
No, only: plugins/ebay.js
Now it should be OK.

PR #519, opened 2020-05-31
Rewrite of Google Images plug-in

GrosPoulet:
This new version of plug-in catches "on the fly" data sent by Google servers when user scrolls down images results.
Note: this method could be applied to other plug-ins, removing the need for API calls.

### DIFF
--- a/plugins/google.js
+++ b/plugins/google.js
@@ -5,6 +5,8 @@ hoverZoomPlugins.push({
     prepareImgLinks:function (callback) {
         var res = [];
         var initData = null;
+        var hookedData = sessionStorage.getItem('hookedData');
+        sessionStorage.removeItem('hookedData');
 
         // Google+ full page viewer
         if (location.search.indexOf('pid=') > -1) {
@@ -28,28 +30,25 @@ hoverZoomPlugins.push({
         );
 
         // Hook Google 'Open' XMLHttpRequests to catch data & metadata associated with pictures displayed
-        // These requests are issued by client to Google servers in order to obtain new data when user scrolls down
-        // Hooked data is stored in scripts in DOM with class='hoverZoomHookedData'
+        // These requests are issued by client side to Google servers in order to obtain new data when user scrolls down
+        // Hooked data is stored in sessionStorage
         if ($('script.hoverZoomHook').length == 0) { // Inject hook script in document if not already there
             var hookScript = document.createElement('script');
             hookScript.type = 'text/javascript';
             hookScript.text = `if (typeof oldXHROpen !== 'function') { // Hook only once!
-                console.log('Hooking XHROpen');
                 oldXHROpen = window.XMLHttpRequest.prototype.open;
                 window.XMLHttpRequest.prototype.open = function(method, url, async, user, password) {
                     // catch responses
                     this.addEventListener('load', function() {
                         try {
-                            //console.log('load: ' + this.responseText);
-                            // store response as plain text in a <script> element for later usage by plug-in
-                            var hookedDataScript = document.createElement('script');
-                            hookedDataScript.type = 'text';
-                            hookedDataScript.text = this.responseText;
-                            hookedDataScript.classList.add('hoverZoomHookedData'); // add specific class for easier retrieval
-                            (document.head || document.documentElement).appendChild(hookedDataScript);
+                            // store response as plain text in a sessionStorage for later usage by plug-in
+                            let storedHookedData = sessionStorage.getItem('hookedData');
+                            if (storedHookedData == undefined) sessionStorage.setItem('hookedData', this.responseText);
+                            else sessionStorage.setItem('hookedData', storedHookedData + this.responseText);
+
                             // Add & remove empty <a> element to/from DOM to trigger HoverZoom,
                             // so data & metadata just added to DOM in <script> element can be exploited
-                            var fakeA = document.createElement('a');
+                            let fakeA = document.createElement('a');
                             (document.head || document.documentElement).appendChild(fakeA);
                             (document.head || document.documentElement).removeChild(fakeA);
                         } catch {}
@@ -66,7 +65,9 @@ hoverZoomPlugins.push({
             if (url != undefined) {
                 url = url.replace(/\\\\/g, '\\');
                 url = url.replace(/\\[uU]00/g, '%');
-                url = decodeURIComponent(decodeURIComponent(url));
+                try {
+                    url = decodeURIComponent(decodeURIComponent(url));
+                } catch {}
             }
             return url;
         }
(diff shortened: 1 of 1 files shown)


plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
-        // Extract direct links to images using document.scripts data (no api call(s) needed)
-        // Mostly inspired from chocolatebot's userscript : Google DWIMages 
+        // Hook Google 'Open' XMLHttpRequests to catch data & metadata associated with pictures displayed
+        // These requests are issued by client to Google servers in order to obtain new data when user scrolls down
+        // Hooked data is stored in scripts in DOM with class='hoverZoomHookedData'
+        if ($('script.hoverZoomHook').length == 0) { // Inject hook script in document if not already there
+            var hookScript = document.createElement('script');
+            hookScript.type = 'text/javascript';
+            hookScript.text = `if (typeof oldXHROpen !== 'function') { // Hook only once!
+                console.log('Hooking XHROpen');
```
extesy, line 37: Please don't pollute console with log messages. You should use `cLog` instead (defined in hoverzoom.js).

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
-        function parseScripts() {
-            scripts = Array.from(document.scripts);
-            callbacks = scripts.filter(script => /^AF_initDataCallback\b/.test(script.text));
-            dataFromScripts = callbacks.pop().text;
-        };
 
-        parseScripts();
+        // Extract data in initDataCallback
+        function extractInitData(tbnid) {
+            console.log('extractInitData');
```
extesy, line 86: cLog

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
+                let lastquoteIndex = initData.indexOf('"', firstquoteIndex + 1);
+                url = initData.substring(firstquoteIndex + 1, lastquoteIndex);
+            }
+            url = cleanUrl(url);
+            return url;
+        }
+
+        // Extract hooked data stored in <script> elements appended to DOM
+        function extractHookedData(tbnid) {
+            console.log('extractHookedData');
```
extesy, line 107: cLog

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
+                            (document.head || document.documentElement).appendChild(fakeA);
+                            (document.head || document.documentElement).removeChild(fakeA);
+                        } catch {}
+                    });
+                    // Proceed with original function
+                    return oldXHROpen.apply(this, arguments);
+                }
+            }`;
+            hookScript.classList.add('hoverZoomHook');
+            (document.head || document.documentElement).appendChild(hookScript);
```
extesy, line 62: I don't understand why do you need to dynamically create a script. All scripts are already executed by the browser, such as for example this `plugins/google.js` file.

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
+                    this.addEventListener('load', function() {
+                        try {
+                            //console.log('load: ' + this.responseText);
+                            // store response as plain text in a <script> element for later usage by plug-in
+                            var hookedDataScript = document.createElement('script');
+                            hookedDataScript.type = 'text';
+                            hookedDataScript.text = this.responseText;
+                            hookedDataScript.classList.add('hoverZoomHookedData'); // add specific class for easier retrieval
+                            (document.head || document.documentElement).appendChild(hookedDataScript);
+                            // Add & remove empty <a> element to/from DOM to trigger HoverZoom,
```
extesy, line 50: Instead of doing that hack, why not call appropriate hoverzoom function directly?

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
+            hookScript.type = 'text/javascript';
+            hookScript.text = `if (typeof oldXHROpen !== 'function') { // Hook only once!
+                console.log('Hooking XHROpen');
+                oldXHROpen = window.XMLHttpRequest.prototype.open;
+                window.XMLHttpRequest.prototype.open = function(method, url, async, user, password) {
+                    // catch responses
+                    this.addEventListener('load', function() {
+                        try {
+                            //console.log('load: ' + this.responseText);
+                            // store response as plain text in a <script> element for later usage by plug-in
```
extesy, line 44: Why do you need to store it as script instead of using a `data-` attribute?

plugins/google.js
```diff
@@ -26,35 +27,132 @@ hoverZoomPlugins.push({
...
+                initData = goodScripts.pop().text;
+            }
+            let tbnidq = '"' + tbnid + '",';
+            if (initData.indexOf(tbnidq) != -1) {
+                let firstquoteIndex = initData.indexOf('[', initData.indexOf('[', initData.indexOf(tbnidq)) + 1) + 1;
+                let lastquoteIndex = initData.indexOf('"', firstquoteIndex + 1);
+                url = initData.substring(firstquoteIndex + 1, lastquoteIndex);
+            }
+            url = cleanUrl(url);
+            return url;
```
extesy, line 102: These two lines can be combined.

plugins/google.js
GrosPoulet: ok, done

plugins/google.js
GrosPoulet: ok, done

plugins/google.js
GrosPoulet: ok, done

plugins/google.js
GrosPoulet: ok, done

plugins/google.js
GrosPoulet: It's a script with type = 'text', so it'll not be executed by browser.
It's just a place to store chunks of data returned by google when user scrolls down results.

plugins/google.js
GrosPoulet: Yes, it's a hack to trigger the function: prepareImgLinksAsync(dontResetDelay)
I can change it, but this function would have to become public ?

plugins/google.js
GrosPoulet: Some chunks of data to store are quite bigs (around 30/50 Ko), so i don't know if datasets are appropriate... 
What's your opinion?

extesy: You can also use a local (preferably) or global variable, similar to existing `hoverZoomPlugins` and `hoverZoom`. That would be even better.

plugins/google.js
extesy: I think calling prepareImgLinksAsync would be better. It's ok to make it public.

plugins/google.js
extesy: Javascript variables can hold state until the page is closed, you don't need to use DOM to store data.

GrosPoulet: I modified this part: XHR responses are now stored in sessionStorage. So no more modifications to DOM needed.

plugins/google.js
GrosPoulet: Unfortunatly, the code that intercepts XHR responses is executed <b>in the page context</b>, not in the extension context. So i did not manage to call prepareImgLinksAsync from here, maybe you have a solution?

plugins/google.js
GrosPoulet: I now use sessionStorage, no need for dataset updates or variables.

GrosPoulet:
No, i was waiting for your advices on lines 47-50

plugins/google.js
extesy: Hm, ok, let's keep it as-is for now and maybe we'll come up with a less hacky approach.

extesy:
Ah, sorry, I don't have any good ideas right now so we can ignore that comment.

PR #522, opened 2020-06-18
Instagram plug-in rewrite

GrosPoulet:
This plug-in is similar to updated Goggle plug-in: it takes advantage of XHR data sent by server when user scrolls down.
There are also **2 helper functions** (getKeysInObject & getValuesInObject) **to extract key(s)/value(s) from any JSON object**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Firefox and Google Chrome extension for zooming images on mouse hover.
 
 ## Download Links
 * [Chrome Web Store](https://chrome.google.com/webstore/detail/hover-zoom%20/pccckmaobkjjboncdfnnofkonhgpceea)
-* [Mozilla Add-ons](https://addons.mozilla.org/en-US/firefox/addon/hover-zoom2/)
+* [Mozilla Add-ons](https://addons.mozilla.org/en-US/firefox/addon/hover-zoom2/) - supported by [@fhadiel](https://github.com/fhadiel) in [another repo](https://github.com/fhadiel/hoverzoom), please report Firefox issues in that repo.
 ---
 
 > Please help with [localizing this extension](https://crowdin.com/project/hoverzoom) to the language you are familiar with!
--- a/_locales/bg/messages.json
+++ b/_locales/bg/messages.json
@@ -160,7 +160,7 @@
         "description": "[options] Zoom videos (WebM, MP4) option"
     },
     "optZoomVideosTooltip": {
-        "message": "Zoom videos (WebM, MP4)",
+        "message": "Увеличение на видеоклипове (за WebM, MP4)",
         "description": "[options] Tooltip for zoom videos (WebM, MP4) option"
     },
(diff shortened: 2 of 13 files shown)


extesy:
@GrosPoulet I think it can be done much easier in case of instagram because it has a public api which works using json. After some googling for `instagram graph "query_hash"` (which you can see in the list of network requests) I found https://github.com/realsirjoe/instagram-scraper/blob/master/igramscraper/endpoints.py which led me to queries like `https://www.instagram.com/p/%s/?__a=1`. So when you have a post, for example https://www.instagram.com/p/CBi07RNsHI6/ then you can append `?__a=1` to it like https://www.instagram.com/p/CBi07RNsHI6/?__a=1 and it will return valid json that you can easily parse without resorting to XHR tricks.

GrosPoulet:
I know there is a public API, look at line 260 :)
I prefer to use this strategy:
1. search url into init data (data found on the page in window._sharedData)
2. if not found then search into data from XHR requests
3. if still not found then proceed with API call
This way, i do not have to wait for an API response and the day this API becomes private methods 1 and 2 might still work.

extesy:
Ok, I'll accept it for now but I feel that this method is very sensitive to future Chrome updates: they keep blocking off more and more functionality from extensions in the name of privacy and security. Can't blame them, really, but they had crippled uBlock Origin already and XHR intercept might be next.

extesy:
@GrosPoulet Actually, now I've tried it locally an the plugin doesn't work for me. I can't zoom neither the main feed https://www.instagram.com nor any pages like https://www.instagram.com/brandonwoelfel/

GrosPoulet:
That's weird, for me https://www.instagram.com/brandonwoelfel/ works.
Chrome Version: 83.0.4103.106 (Build officiel) (64 bits)

GrosPoulet:
Very few images do not work on Google images, can you open the console and see what's wrong ?

extesy:
I'm also on 83.0.4103.106. Do you have anything enabled on about:flags page? For instagram there's nothing on console besides logging requests blocked by uBlock. No errors that I can see.

GrosPoulet:
I did **Reset All** in chrome://flags
I have just one extension loaded (unpacked): HoverZoom+
I'll install uBlock and check again, maybe you can desactivate it and see what happens

GrosPoulet:
I installed ublock Origin 1.27.10: still no problem.

extesy:
I found a problem: you have renamed plugin from `instagram` to `Instagram` so it became automatically disabled in the list of active plugins.

extesy:
Maybe make plugin settings case insensitive?

GrosPoulet:
That's weird because even if i change: **name:'instagram'** to : **name:'Instagram'** or even: **name:'123Instagram'** it still works.
The name is only used in Options to disable/enable a specific plug-in.

PR #543, opened 2020-08-02
Typos in traductions (Fr + Es)

GrosPoulet:


extesy:
@GrosPoulet This will all be overwritten by crowdin files, you should fix it over there instead: https://crowdin.com/project/hoverzoom

PR #628, opened 2020-10-24
New plugin: Goodreads

dubhater:
Hi!

This loads book covers and user, author, and group profile images.


extesy:
Nice, thank you!

PR #630, opened 2020-11-09
Fix e621/e926 unreliable image hovering

pinnae:
Original image URL is contained in the data-file-url attribute so use it
instead of guessing. Hovering on WebM files works now and only one
image is requested


extesy:
@pinnae Thank you for the contribution!

PR #643, opened 2020-12-31
New colors for image loader + fix on urls normalization

GrosPoulet:
- image loader colors saturation reduced to 25%
- correct handling of urls of type: blob:http...
They are not very common but can be find on whatsapp:  blob:https://web.whatsapp.com/fb375aba-64bd-4d8e-89c6-dc85700e79c2

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -500,7 +500,7 @@ var hoverZoom = {
             }
 
             var src = data[hoverZoomSrcIndex];
-			if (src && src.indexOf('http:') !== 0 && src.indexOf('https:') !== 0 && src.indexOf('file:') !== 0 && src.indexOf('blob:http:') !== 0 && src.indexOf('blob:https:') !== 0) {
+            if (src && src.indexOf('http:') !== 0 && src.indexOf('https:') !== 0 && src.indexOf('file:') !== 0 && src.indexOf('blob:http:') !== 0 && src.indexOf('blob:https:') !== 0) {
                 if (src.indexOf('//') !== 0) {
                     if (src.indexOf('/') !== 0) {
                         // Image has relative path (doesn't start with '/')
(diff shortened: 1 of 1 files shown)


js/hoverzoom.js
```diff
@@ -500,7 +500,7 @@ var hoverZoom = {
             }
 
             var src = data[hoverZoomSrcIndex];
-            if (src && src.indexOf('http') !== 0) {
+            if (src && src.indexOf('http:') !== 0 && src.indexOf('https:') !== 0 && src.indexOf('file:') !== 0 && src.indexOf('blob:http:') !== 0 && src.indexOf('blob:https:') !== 0) {
```
extesy, line 503: Should it be `||` instead of `&&`? How can all of these strings be present at the same time?

GrosPoulet: No, this function looks for too short url: url that does **not** start with http:, https:, and so on.
If url is too short then it normalizes url by prepending missing information (using windows.location.protocol, host, and pathname).

For instance: _/pictures/01.jpg_ 
will be changed to:  _https://www.mywebsite.com/pictures/01.jpg_

GrosPoulet: Are you ok with new colors scheme? Quite similar to Imagus' btw...

extesy: Ah, I see what you mean. In this case I think the intention would be more obvious if instead of `!== 0` it used `=== -1` since -1 means "not found".

extesy: I haven't tried the new color scheme yet, but 25% saturation seems about right to me.

GrosPoulet: - _=== -1_ checks that src does not contain _http:_ **at any location**
- _!== 0_ checks that src does not **start** with _http:_

An url can contain another url, like this one: 
_https://imageproxy.pimg.tw/resize?url=https://live.staticflickr.com/65535/49745339392_2ebe339af0_b.jpg&maxwidth=530&maxheight=530_
so in this case _=== -1_ (not found) is not enough.

extesy: Good point.

PR #667, opened 2021-01-24
Fix for 2 plug-ins + 7 new plug-ins

GrosPoulet:
Fix for plug-ins:
- 500px
- Flickr

New plug-ins:
- Aminus3
- AOL
- Derpibooru
- Giphy
- Imageevent
- Pikwizard
- Thingiverse

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -90,7 +90,8 @@
         },
         {
             "js": ["plugins/flickr.js"],
-            "matches": ["*://*.flickr.com/*"]
+            "matches": ["*://*.flickr.com/*",
+                "*://*.flickr.net/*"]
         },
         {
             "js": ["plugins/yahoo.js"],
@@ -1333,7 +1334,7 @@
         },
         {
             "js": ["plugins/stocksnap.js"],
-            "matches": ["*://*.stocksnap.io/*"]
+            "matches": ["*://*.stocksnap.io/*", "*://*.stocksnap.com/*"]
         },
         {
             "js": ["plugins/pexels.js"],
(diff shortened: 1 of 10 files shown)


extesy:
@GrosPoulet thingiverse plugin doesn't work for me, I've tried home page and it converts urls like https://cdn.thingiverse.com/assets/94/65/b7/2d/da/card_preview_Cierre_Clamp_V2.png into https://cdn.thingiverse.com/assets/94/65/b7/2d/da/card_display_large.png which does not exist.

extesy:
Flickr is also a bit weird. On some images on the first hover is shows a "medium" image and on the second hover it shows a "large" one. And I can't reliably reproduce that. This effect seems to appear on random images on https://www.flickr.com/search/?text=nature page.

GrosPoulet:
@extesy 
- Thingiverse plug-in is not perfect, only some full urls are guessed correctly. I need some more inspiration!
- Flickr plug-in performs some async API calls that may return results (= fullsize urls) **after** medium urls have been displayed. (Medium urls computation require no API calls, simple regex)

Thanks a lot for testing these plug-ins, did you try the one for 500px?

extesy:
Yes, I've tried 500px and seems to work great, thank you! As for the thingiverse, I haven't been able to find any single thumbnail which worked so I would prefer to not include it at this stage yet.

GrosPoulet:
The Thingivers plug-in works well here: https://www.thingiverse.com/education
...but not so well elsewhere, so maybe you should remove it or open an issue so someone can try to fix it (could be me!)

extesy:
Here are three different sizes of thingiverse urls:
* https://cdn.thingiverse.com/assets/94/65/b7/2d/da/card_preview_Cierre_Clamp_V2.png
* https://cdn.thingiverse.com/assets/2b/7f/d0/7d/c9/featured_preview_Cierre_Clamp_V2.png
* https://cdn.thingiverse.com/assets/47/fd/d6/9d/5a/large_display_Cierre_Clamp_V2.png

As you can see, asset ids are different for each of them so I don't think a simple regex replace would help here. Thingiverse would need a `prepareFromDocument` to fetch the images. I can also see a call to `https://api.thingiverse.com/things/4699910/images` made which has all the information needed to display albums and big images: https://pastebin.com/D6czJkXT

GrosPoulet:
Yes, an API call would be the best solution. I'll try to give it a shot next week.

PR #671, opened 2021-01-28
Fix for red spinner: now not displayed by default

GrosPoulet:
@extesy, i first decided to remove all spinners (green, orange, red) but i changed my mind and kept the green one. 
I spared it because it's the only visual clue for users that HZ+ is actually doing something (=trying to zoom the image under mouse pointer!).
Another option whould be to modify the look of image being hovered, like Imagus that adds a dotted border around image.
I did some tests, it is not fully reliable, so maybe we should keep the green spinner for now, what do you think?
By default, red and future orange spinners are disabled.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -367,6 +367,14 @@
         "message": "%",
         "description": "[options] Unit name (percentage) for opacity option"
     },
+    "optDisplayImageLoaderTooltip": {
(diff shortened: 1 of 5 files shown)


extesy:
I think there's a little conflict between the option description and it's tooltip. Can you please clarify what exactly this option (`optDisplayImageLoader`) is doing? Does it completely disable any spinner or just the error one? The message says "display image loader" but the tooltip says "orange & red spinners are optional". It almost sounds like the option controls the whole spinner thingy and there should be a separate option just for error/orange ones.

PR #688, opened 2021-02-13
HZ+ now works on Gelbooru.com

### LINKED ISSUE #61: Hover functionality for gelbooru.com
Image hover does not work on http://gelbooru.com/

GrosPoulet:
This update closes #61 (an issue opened in 2016 :open_mouth:)

### DIFF
--- a/plugins/safebooru.js
+++ b/plugins/safebooru.js
@@ -1,14 +1,14 @@
 ﻿var hoverZoomPlugins = hoverZoomPlugins || [];
 hoverZoomPlugins.push({
     name:'Safebooru',
-    version:'0.1',
+    version:'0.2',
     prepareImgLinks:function (callback) {
         var res = [];
 
         hoverZoom.urlReplace(res,
             'img[src]',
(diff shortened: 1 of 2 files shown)


plugins/safebooru.js
```diff
@@ -19,10 +19,10 @@ hoverZoomPlugins.push({
...
         hoverZoom.urlReplace(res,
             'img[src]',
-            /(.*)\/(?:thumbnails|samples)\/(.*)\/(?:thumbnail|sample)_(.*)\.jpg(.*)/,
-            '$1/images/$2/$3.png'
+            /(.*)\/(?:thumbnails|samples)\/(.*)\/(?:thumbnail|sample)_(.*)/,
+            '$1/images/$2/$3'
         );
 
-        callback($(res));
+        callback($(res), this.name);
```
extesy, line 26: I'm just curious, what is this doing? The callback function `imgLinksPrepared` only takes one parameter.

GrosPoulet: It's for debugging, so you know which plugin found the high resolution url (it is not always obvious).
I use a slightly different version of HZ+where this info is handled.

extesy: I see, thank you.

PR #739, opened 2021-06-10
Fix img elements title's not being removed

### LINKED ISSUE #252: Alt/Title text obscuring image
With a recent update (probably March 21, 2017) images with alt text or title attributes in the image element will cause the popup to be layered in front of the hover-zoomed image. This results in the image being obscured by the popup. 

With some research, it appears that there is no way to disable alt text/ titles in Google Chrome. There also appears to be no way to disable alt text/ titles in the options for the extension.

Example image: http://imgur.com/a/nATt7

Perhaps creating a quick toggle in the settings that enables/ disables displaying the alt would be a quick solution.

### LINKED ISSUE #731: Feature request: Disable image titles when viewing image
**Is your feature request related to a problem? Please describe.**
When trying to view images, if the image has a title attribute, this will show up blocking some of the image.

**Describe the solution you'd like**
Either disabling the title attribute showing on hover, or make the image appear in front.

**Additional context**
Help
The image title (added from inspect element) appears over the image. Is there anyway to disable this either browser wide, or with Hover Zoom?

Mattwmaster58:
fixes #731

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -449,8 +449,8 @@ var hoverZoom = {
             }
             titledElements = img.parents('[title]').not('iframe, .lightbox, [rel^="lightbox"]');
             // Occasionally, the img element itself will have a title attribute
-            // which we like to  handle that as well
-            titledElements = titledElements.add(img.filter('[title]');
+            // which we like to handle that as well
+            titledElements = titledElements.add(img.filter('[title]'));
             titledElements.each(function () {
                 $(this).data().hoverZoomTitle = this.getAttribute('title');
                 this.title = '';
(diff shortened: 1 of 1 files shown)


js/hoverzoom.js
```diff
@@ -448,6 +448,9 @@ var hoverZoom = {
                 restoreTitles();
             }
             titledElements = img.parents('[title]').not('iframe, .lightbox, [rel^="lightbox"]');
+            // Occasionally, the img element itself will have a title attribute
+            // which we like to  handle that as well
+            titledElements = titledElements.add(img.filter('[title]');
```
extesy, line 453: @Mattwmaster58 Thank you for the pull request. Did you actually test it? You have a syntax error on this line.

Mattwmaster58: I swear I had this working, then all of a sudden the extension wasn't loading...I figured firefox was doing something strange but this would explain it LOL

PR #746, opened 2021-06-24
Initial prototype for locking image on screen, allowing user to zoom and pan around.

EhsanKia:
This tries to imitate the image locking behavior from Imagus. By default it's bound to L, but can also be set to right click to match the behavior on Imagus. Although maybe it should be bound to None by default, let me know what you think.

The PR doesn't update all the localization files, not sure if there's an automated way of doing that, but I can do that later in the review.

The change itself should have nearly no impact on the default behavior, from what I can tell. I do need to test it some more with edge cases and different settings. It is a big of a large change and touches quite a lot, but most of it is gated behind imageLocked which can only be set true by using this specific feature. The behavior with imageLocked = false should be identical from what I can tell.

Let me know what you think.


GrosPoulet (commented):
@extesy, @EhsanKia : may i commit my changes first please (https://github.com/extesy/hoverzoom/projects/1#card-62121594) ?
I hope i'll be done by the end of this week.
BTW, i would appreciate some feedback/review/tests too, since i made large & impacting changes :mask:
Since i rewrote some parts @EhsanKia worked on, merge will be non-trivial : we can handle it later on.

EhsanKia:
Sure I don't mind rebasing on top

extesy:
@EhsanKia I've tried this PR and it seems to work, the only issue I've noticed so far is that the default `L` shortcut conflicts with the very popular Reddit Enhancement Suite (aka RES) extension. Perhaps leave it unassigned or find some key which doesn't conflict with RES?

EhsanKia:
Good call, I didn't test it much with L myself. Set default to None. Do you have any tip for generating/populating the i18n files? Otherwise I might do another cleanup pass and wait for @GrosPoulet's PR to rebase.

extesy:
The rest of i18n will be populated by crowdin integration, you only need to update a single English file, which you already did.

GrosPoulet:
@EhsanKia , @extesy : i finally commited my changes, sorry for the delay.
I did a fair amount of testing, but do not hesitate to check it :microscope:

EhsanKia:
Thanks for the heads up, I should have time to rebase this weekend. Doesn't seem too bad from a quick look.

GrosPoulet:
I solved conflicts, i think we're ready for merging.

extesy:
Thank you! I'll test drive both commits for a few more days and then will publish a new version.

js/hoverzoom.js
```diff
@@ -340,8 +363,11 @@ var hoverZoom = {
                 //hz.hzImg.css('visibility', 'visible');
 
                 // image natural dimensions
-                imgDetails.naturalWidth = imgFullSize.width() * zoomFactor;
-                imgDetails.naturalHeight = imgFullSize.height() * zoomFactor;
+                //imgDetails.naturalWidth = imgFullSize.width() * options.zoomFactor;
+                //imgDetails.naturalHeight = imgFullSize.height() * options.zoomFactor;
+                imgDetails.naturalWidth = (imgFullSize[0].naturalWidth ? imgFullSize[0].naturalWidth : imgFullSize.width()) * options.zoomFactor;
+                imgDetails.naturalHeight = (imgFullSize[0].naturalHeight ? imgFullSize[0].naturalHeight : imgFullSize.height()) * options.zoomFactor;
```
EhsanKia, line 369: The only part I'm not sure about is here. I think you gotta use zoomFactor and not options.zoomFactor.
I explicitly take a snapshot of options.zoomFactor at the the top of loadFullSizeImage and then later modify it inside documentOnMouseWheel. This will break mousewheel zooming of the locked image I believe

EhsanKia:
Bump, I just tried the new released version and the zoom functionality is indeed broken due to the bad merge above. Do you want me to submit a PR? Just need to change `options.zoomFactor` for `zoomFactor` on both those lines.

PR #751, opened 2021-07-08
New image/video viewer

GrosPoulet:
- new option: display image/video details (dimensions, filename...) above or below image/video (default: disabled)
- new option: font size for caption & details (default: 11px)
- new option: font outline for caption & details (default: disabled)
- new option: frame thickness (default: 4px)
- gallery index (1/3, 2/3, ...) now displayed in top-left corner of image/video (so it does not mix with timestamp in case of video)
- improvements to video display
- bugs fixed
- plugins updated

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -111,6 +111,34 @@
         "message": "None",
         "description": "[options] Caption location option to disable captions"
     },
+    "optDetailsLocation": {
+        "message": "Details position:",
+        "description": "[options] Adjust the position of details"
+    },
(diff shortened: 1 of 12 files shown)


extesy:
Looks great so far! I haven't noticed any bugs yet. Image/video frame is new so the zoomed image looks a bit different with it. I still like it, but with 350K+ users there will always be some like "who moved my cheese?" so we should check user feedback on that.

extesy:
I'm going to merge your PR to help resolve conflicts with the other pull requests. If I find any issues, we'll fix it forward.

PR #767, opened 2021-07-29
Image lock : added pan around for videos

GrosPoulet:


GrosPoulet:
@EhsanKia i added pan around functionality for videos too.
It would be nice to add controls (play, pause,...) to videos, as in Imagus, i'll try to give it a shot since i did something similar when a  video is opened  in a separate window (W action key).

EhsanKia:
I don't fully recall how it was on Imagus since it's been a while, but is it not as simple as settings controls on the <video> element? I actually usually do it manually by rick clicking the (locked) video and checking "enable video controls" which does the same thing.

Issue #765, opened 2021-07-28
Copy URL of image or video


RickCarletti:
It would be nice to be able to copy the url of the image or video when opening the popup content.

When used on 9gag I found the need to copy the content url.

extesy:
Fixed by https://github.com/extesy/hoverzoom/commit/9bf7dd42720e16b125549fa80e24ebe207fd2f1b

Issue #750, opened 2021-07-08
Add copy to clipboard hotkey


kazimirr:
There's a hotkey for saving images quickly. It would be nice to have another hotkey to quickly copy an image to the clipboard.

Use case:
I have a group chat with my friends in Telegram. We often share pictures with each other. With such hotkey sharing would be quicker.

extesy:
Fixed by https://github.com/extesy/hoverzoom/commit/9bf7dd42720e16b125549fa80e24ebe207fd2f1b

Issue #459, opened 2019-09-17
[Feature Request] shortcut copy image


bnpl7:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Steps to reproduce the problem (include image url here)
1.
2.
3.

62274677:
I would like to bump this because I too would like this feature and came here to request it!

PR #768, opened 2021-07-31
Feat: Add Ability to Copy Images and Image URLs into the Clipboard

George-lewis:
Hello maintainers of HoverZoom, I raise this PR to add some long requested functionality: The ability to copy images and their links into the clipboard.

This should address the following issues: #459, #750, and #765 

I have set the default keys to C for "<ins>**C**</ins>opy" and L for "Copy <ins>**L**</ins>ink".

The functionality is fairly straightforward, you hover over an image and press one of the buttons to either copy the image into your clipboard, or the direct link to that image or video.

I tried to add entries to all the English locales I could think of, please let me know if I missed any.

Unfortunately there's an unavoidable caveat to the copy image functionality. Some sites have their CORS set up in such a way (I have found a few) that while you can hoverzoom an image on a particular page, you cannot interact with it in the script. My current code will emit an error when calling `.toBlob`, and it will complain about attempting to export a [tainted canvas](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image). Similarly you cannot fetch the source URL either. 

Let me know what you think. Thanks.

### DIFF
--- a/_locales/ca/messages.json
+++ b/_locales/ca/messages.json
@@ -547,22 +547,6 @@
         "message": "When this key is held down, the picture is displayed using all available space. Useful for high resolution pictures only.",
         "description": "[options] Action key description"
     },
-    "optCopyImageKeyTitle": {
-        "message": "Copy image to clipboard",
-        "description": "[options] Action key title"
-    },
-    "optCopyImageKeyDescription": {
-        "message": "When this key is pressed, the image is copied into the clipboard.",
-        "description": "[options] Action key description"
-    },
-    "optCopyImageUrlKeyTitle": {
-        "message": "Copy image URL to clipboard",
-        "description": "[options] Action key title"
-    },
-    "optCopyImageUrlKeyDescription": {
-        "message": "When this key is pressed, the image's or video's URL is copied into the clipboard.",
-        "description": "[options] Action key description"
-    },
     "optPrevImgKeyTitle": {
         "message": "View previous image in a gallery",
         "description": "[options] Action key title"
--- a/_locales/ca/messages.json
+++ b/_locales/ca/messages.json
@@ -547,6 +547,22 @@
         "message": "When this key is held down, the picture is displayed using all available space. Useful for high resolution pictures only.",
         "description": "[options] Action key description"
     },
+    "optCopyImageKeyTitle": {
+        "message": "Copy image to clipboard",
+        "description": "[options] Action key title"
+    },
+    "optCopyImageKeyDescription": {
+        "message": "When this key is pressed, the image is copied into the clipboard.",
+        "description": "[options] Action key description"
+    },
+    "optCopyImageUrlKeyTitle": {
+        "message": "Copy image URL to clipboard",
+        "description": "[options] Action key title"
+    },
+    "optCopyImageUrlKeyDescription": {
+        "message": "When this key is pressed, the image's or video's URL is copied into the clipboard.",
+        "description": "[options] Action key description"
+    },
     "optPrevImgKeyTitle": {
         "message": "View previous image in a gallery",
         "description": "[options] Action key title"
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2111,6 +2111,8 @@ var hoverZoom = {
         }
 
         function copyImage() {
+            if (!isChromiumBased) return;
+
             const url = imgDetails.url;
             if(!url) return;
 
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1678,9 +1678,11 @@ var hoverZoom = {
                     saveImage();
                     return false;
                 }
-                if (event.which == options.copyImageKey) {
-                    copyImage();
-                    return false;
+                if (isChromiumBased) {
+                    if (event.which == options.copyImageKey) {
+                        copyImage();
+                        return false;
+                    }
                 }
                 if (event.which == options.copyImageUrlKey) {
                     copyLink();
@@ -2111,8 +2113,6 @@ var hoverZoom = {
         }
 
         function copyImage() {
-            if (!isChromiumBased) return;
-
             const url = imgDetails.url;
             if(!url) return;
 
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2103,26 +2103,13 @@ var hoverZoom = {
             saveAudio();
         }
 
-        function _copyImage(url){
-            var img=document.createElement('img');
-            img.src=url;
-            document.body.appendChild(img);
-            var r = document.createRange();
-            r.setStartBefore(img);
-            r.setEndAfter(img);
-            r.selectNode(img);
-            var sel = window.getSelection();
-            sel.addRange(r);
-            document.execCommand('Copy');
-        }
-
         function copyLink() {
             if (!hz.hzImg) return;
             let img = hz.hzImg.find('img').get(0);
             let video = hz.hzImg.find('video').get(0);
             let target = img || video;
             if (!target) return;
-            
+
             let src = target.src;
 
             navigator.clipboard.writeText(src);
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2140,8 +2140,7 @@ var hoverZoom = {
         }
 
         function copyImage() {
-            if (!hz.hzImg) return;
-            let img = hz.hzImg.find('img').get(0);
+            const img = hz.hzImg?.find('img')?.get(0);
             if (!img) return;
 
             toBlob(img)
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1678,6 +1678,14 @@ var hoverZoom = {
                     saveImage();
                     return false;
                 }
+                if (event.which == options.copyImageKey) {
+                    copyImage();
+                    return false;
+                }
+                if (event.which == options.copyImageUrlKey) {
+                    copyLink();
+                    return false;
+                }
                 // Cancels event if an action key is held down (auto repeat may trigger additional events)
                 if (event.which == options.actionKey ||
                     event.which == options.fullZoomKey ||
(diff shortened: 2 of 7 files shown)


js/hoverzoom.js
```diff
@@ -2095,6 +2103,52 @@ var hoverZoom = {
...
+        async function toBlob(img) {
+            const isPng = img.src?.endsWith('png');
+
+            // Fetch cached image data if it's a png
+            if (isPng) {
+                return fetch(img.src).then(res => res.blob())
+            }
+
+            // Else render to a canvas to convert to a png
+            const canvas = document.createElement("canvas");
```
extesy, line 2127: Is it possible to use `imgDetails.url` to copy the actual image file like https://web.dev/async-clipboard/ article did? It might resolve some of your problems with canvas.

_locales/ca/messages.json
```diff
@@ -547,6 +547,22 @@
         "message": "When this key is held down, the picture is displayed using all available space. Useful for high resolution pictures only.",
         "description": "[options] Action key description"
     },
+    "optCopyImageKeyTitle": {
```
extesy, line 550: There's no need to update any localization files other than `en`, it will be taken care of automatically through https://crowdin.com/project/hoverzoom integration.

extesy (commented):
Thank you for the pull request, @George-lewis! It looks good to me overall, just one question about using alternative to copying canvas.

_locales/ca/messages.json
George-lewis: Ah that makes sense. I've reverted the `ca` and `uk` locale files.

js/hoverzoom.js
George-lewis: I'm not too familiar with all the stuff going on with the global variables in this file. Is `imgDetails.url` the same as `hz.hzImg.find('img').get(0).src`?

I'm having trouble finding docs for this, but I get the impression that Chrome doesn't support copying non-png images.
I think [this source file](https://github.com/chromium/chromium/blob/d7da0240cae77824d1eda25745c4022757499131/third_party/blink/renderer/core/clipboard/clipboard_mime_types.cc) documents all of the supported types in the clipboard, but I'm not sure.

In any case when I try to copy a jpeg the browser says: `Uncaught (in promise) DOMException: Type image/jpeg not supported on write.` Feel free to confirm this for yourself, you could modify `toBlob` to always do a fetch, and then try to copy a jpeg. This is why I'm using the canvas, it's the recommended way to convert images to png in the browser. From your article:

> To write an image to the clipboard, you need the image as a blob. One way to do this is by requesting the image from a server using fetch(), then calling blob() on the response.
>
> Requesting an image from the server may not be desirable or possible for a variety of reasons. Fortunately, you can also draw the image to a canvas and call the canvas' toBlob() method.

Thoughts?

extesy: `imgDetails` gets populated with image details (url, width, height) when it's zoomed. `hz.hzImg.find('img').get(0).src` is kinda the same thing but it tries to parse the details of the artificial `hzImg` node that the extension creates for the full image.

As for the article's phrase "requesting an image from the server may not be desirable or possible for a variety of reasons" - it does not apply in our case because we _always_ load the full image from the server, that's kinda the point of this extension. You already do this:

            if (isPng) {
                return fetch(img.src).then(res => res.blob())
            }

What happens if you remove "if (isPng)"? I haven't tried myself but I've seen images saved with incorrect extensions (png vs jpg) and browsers can still render them correctly even though the content doesn't match the extension.

George-lewis: > What happens if you remove "if (isPng)"?

Removing it (and changing the clipboard item's mime type to match: `{[blob.type]: blob}`) results in trying to copy non-png images to the clipboard, which produces the error mentioned in my last comment. Unless there's something strange about my Chrome, you cannot write other image formats to the clipboard.

If it's convenient for you I think you should try running this in your browser:

```js
const url = "https://i.imgur.com/pg4bkej.jpeg";
fetch(url)
    .then(res => res.blob())
    .then(blob => new ClipboardItem({[blob.type]: blob}))
    .then(item => {
        setTimeout(() => {
            navigator.clipboard.write([item])
        }, 3000);
    });
```

You'll need to run it on a page that allows you to make the request to that origin (a new tab should work). Make sure you click on the browser window before the timeout executes, otherwise clipboard access will be denied: `DOMException: Document is not focused.`

Let me know what happens. I get `DOMException: Type image/jpeg not supported on write.`

extesy: This worked for me:
```javascript
const url = "https://i.imgur.com/pg4bkej.jpeg";
fetch(url)
    .then(res => res.blob())
    .then(blob => blob.arrayBuffer())
    .then(data => new ClipboardItem({["image/png"]: new Blob([data], {"type": "image/png"})}))
    .then(item => {
        setTimeout(() => {
            navigator.clipboard.write([item])
        }, 3000);
    });
```

George-lewis: Fascinating. Now that's a hack if I've ever seen one. This works for me as well, but I've honestly no idea how. If I do this, and then try to paste into an application, it pastes as a png, so the conversion must be done somewhere - maybe on the os level? Though I can't confirm that due to lack of docs and no access to another operating system at the moment. Are you also on Windows? I think we should try this on some other systems.

extesy: It works in Chrome on MacOS too (I've just tried) but it does not work on Firefox on any OS. Not because of the hack but because ClipboardItem is disabled by default: https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem. So this would need an explicit check to avoid syntax errors in browser console.

George-lewis: Well how about that. I really wonder how this works. Oh well, if this is what you want then I'll push it onto my branch later. Sound good?

extesy: Try it out to see if you can find any potential problems. I think that would be a better approach over using canvas.

George-lewis: I feel like it's a bit prone to breaking, but I'll toy around with it a bit more. With regards to Firefox, what should we do there? I thought I'd seen Chrome-specific APIs around the codebase so I was under the impression we didn't support it. Anyway I'm going to be away from my PC for the rest of the night most likely, so I'll get back to it tomorrow.

extesy: Take a look at [README.md](https://github.com/extesy/hoverzoom/blob/master/README.md) - there are links to Chrome, Firefox and Edge extension stores there. We support all three, although most of the users are on Chrome. In regards to Firefox, I think we can put a warning "does not work on Firefox" on these two options, and check for `ClipboardItem` being defined before doing anything with clipboard.

George-lewis: Oh I see. Well good news about Firefox, I think we can use [clipboard.setImageData](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/clipboard/setImageData). I know that Edge is chromium-based now, meaning we can use the same logic for Chrome and Edge, right?

extesy: Yes, I expect Chrome and Edge to behave the same way.

George-lewis: I've made a few discoveries:

1. Despite documentation existing for `clipboard.setImageData`, it does not actually exist in FireFox as far as I can tell.
2. FireFox *does* support `ClipboardItem` / `clipboard.write`, but only if a certain flag is enabled
   - `dom.events.asyncClipboard.clipboardItem`
   - https://caniuse.com/mdn-api_clipboard_write
3. FireFox doesn't accept our hacked blob like Chromium does. When you try to write to to the clipboard: `DOMException: Unable to decode blob for 'image/png' as image.`

I feel that this leaves us with a few options:

1. Do not support FireFox for this at all
2. Ask users to enable the flag if they want the feature. We can then selectively enable it via feature detection (like checking if `window.ClipboardItem` is defined)
   - Do we only allow FireFox users to copy png's?
   - Do we go back to the canvas method to ensure it works the same across both browsers?

extesy: Given the low total number of Firefox users, I think it's ok to not support it for now. Can you please also hide these options completely on Firefox instead of showing them but disabling functionality? Something like `if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1)` should work.

extesy:
By the way, I don't see `options.html` file in this PR. How are you exposing the options?

js/hoverzoom.js
George-lewis: > Given the low total number of Firefox users

:smile: Real jab at FireFox there. Especially considering they have [lost 20% / 50 million users over the last year](https://data.firefox.com/dashboard/user-activity).

> By the way, I don't see `options.html` file in this PR. How are you exposing the options?

The extension dynamically creates the HTML here: https://github.com/extesy/hoverzoom/blob/master/js/options.js#L32
It was sufficient to add the option name to the array, and then the text into the locale files.

>  Can you please also hide these options completely on Firefox

Sure. I'll take the easy way out here and add an explicit check to the code that generates the options page. I can also add an explicit check to the code that does the copying? Though it will partially work as-is

George-lewis: Latest commit should do the trick. It uses `navigator.userAgentData`, which is still technically experimental, but is supported in all Chromium browsers so we're good-to-go. It was the best way I could find to check is a browser is Chromium-based.

js/hoverzoom.js
```diff
@@ -1678,6 +1678,14 @@ var hoverZoom = {
                     saveImage();
                     return false;
                 }
+                if (event.which == options.copyImageKey) {
```
extesy, line 1681: I think it's better to put an `if (isChromiumBased)` block around these two statements to avoid processing these shortcuts on Firefox while not showing the options to re-assign and disable them.

js/hoverzoom.js
extesy: Looks almost perfect, just a comment about shortcuts processing.

js/hoverzoom.js
George-lewis: Good catch. This could have definitely prevented "C" from being assigned to a shortcut processed later. Fixed.

George-lewis:
Before merging:
- remove feature check
   - made redundant by chromium check
- reorder check for options page

I'll do it later

extesy:
Thank you for the contribution, @George-lewis!

GrosPoulet:
Hi @George-lewis, welcome aboard :sailboat:
I added a default action key for **Image Lock** feature (#746) created by @ehsankia : **L** (76)
I choosed the letter **L** for **L**ock, but you are using the same letter for **Copy Image URL** action key.
=> do you mind if i change your action key to **U** (85) for **U**RL ?

George-lewis:
> Hi @George-lewis, welcome aboard ⛵
> I added a default action key for **Image Lock** feature (#746) created by @EhsanKia : **L** (76)
> I choosed the letter **L** for **L**ock, but you are using the same letter for **Copy Image URL** action key.
> => do you mind if i change your action key to **U** (85) for **U**RL ?

Thanks? Go right ahead. **U** sounds like a fine replacement

PR #819, opened 2021-11-21
New plug-in for: Twitch.tv (clips, VOD and live)

GrosPoulet:
New plug-in for: **Twitch.tv** (clips, VOD and live)

- all videos (from Twitch and others) now display standard controls (play, pause, seek, volume). User must lock video first so controls can be used.
- standard audio controls also added in case video and audio streams are distinct (like in Bilibili)
- new JS library added: **HLS.js** (https://github.com/video-dev/hls.js/), so browsers that do not support playlists (=M3U8) can play them (such as Chrome).
- playlists URLs can be downloaded, so users can load them in another viewer, such as VLC.
@extesy: i added the library unminified to ease Chrome Webstore review, maybe it is not necessary ?


extesy:
@GrosPoulet Chrome is a little less strict but Firefox requires unminified libraries plus the url to the original code of this library. Can you please share the url for a specific release or a tag which would 100% match the one you've added?

GrosPoulet:
I downloaded latest release (1.1.1) of HLS.js from here: https://github.com/video-dev/hls.js/releases 
This library seems to be very well taken care of (+10k stars), and works like a charm (at least with Twitch) :heart_eyes:

manifest.json
```diff
@@ -50,7 +50,7 @@
     ],
     "content_scripts": [
         {
-            "js": [ "js/libs/jquery-3.2.1.js", "js/libs/contentScriptsAPIBridge.js", "js/tools.js", "js/plugins.js", "plugins/default.js", "js/hoverzoom.js"],
+            "js": [ "js/hls.js", "js/libs/jquery-3.2.1.js", "js/libs/contentScriptsAPIBridge.js", "js/tools.js", "js/plugins.js", "plugins/default.js", "js/hoverzoom.js"],
```
extesy, line 53: This library is kinda large (~1MB) to include it unconditionally on all sites. If only Twitch uses it, why not include it on line 1692 with the twitch plugin?

GrosPoulet:
Right, i fixed it with #821

PR #828, opened 2021-12-05
Fix for plug-in: cloudfront_a

GrosPoulet:


plugins/cloudfront_a.js
```diff
@@ -1,13 +1,13 @@
...
     name:'cloudfront_a',
-    version:'0.1',
+    version:'0.2',
     prepareImgLinks:function (callback) {
         var res = [];
 
         // sample: https://d38hokjm2drjyk.cloudfront.net/?url=nypost.com%2Fwp-content%2Fuploads%2Fsites%2F2%2F2021%2F03%2Fjen-psaki.jpg%3Fquality%3D90%26strip%3Dall%26w%3D1200&w=300&h=190&secure=yes&token=cdb675f5d8e584c6d6233188352417df6750c099
         //      -> nypost.com/wp-content/uploads/sites/2/2021/03/jen-psaki.jpg
-        var reThumb1 = /.*cloudfront.net\/\?url=(.*)\.(jpe?g|gif|png|webp)(.*)/;
+        var reThumb1 = /.*cloudfront.net.*?\?url=(.*)\.(jpe?g|gif|png|webp)(.*)/;
```
extesy, line 10: @GrosPoulet I don't understand `.*?` pattern, can you please clarify? The `*` means "0 or more instances" and the `?` means "0 or 1 instances". They seem redundant when used together.

GrosPoulet: @extesy: 

.*  : **greedy**
.*? : **non-greedy**

For instance, with url = https://d38hokjm2drjyk.cloudfront.net/blah/blah/blah/?url=nypost.com/wp-content/uploads/sites/2/2021/03/jen-psaki.jpg?quality=90

`url.replace(/.*cloudfront.net.*\?(.*)/, '$1') `
returns:  **quality=90**
because regex grabs all characters until **LAST** "?"  

`url.replace(/.*cloudfront.net.*?\?(.*)/, '$1') `
returns: **url=nypost.com/wp-content/uploads/sites/2/2021/03/jen-psaki.jpg?quality=90** 
because regex grabs all characters until **FIRST** "?"

In the case of Cloudfront, it's a little overkill because i use "?url=" to split.
Since there is only one "?url=" in urls, greedy & non-greedy regex will return the same result.
It would be necessary if there was a second occurence of "?url=" since we want to stop at first one.
You can change it for clarity.

extesy: Very interesting, I didn't know about non-greedy modifier. Thank you for a great explanation!

PR #829, opened 2021-12-05
New plug-in for: StartPage search engine

GrosPoulet:


plugins/startpage.js
```diff
@@ -0,0 +1,182 @@
...
+                    }
+                }
+                return keys;
+            }
+        }
+
+        // Return JSON object corresponding to path, without using the Evil eval
+        // path syntax: [key1][key2][key3]...
+        function getJsonObjectFromPath(path) {
+            return new Function('return ' + spJson + path)();
```
extesy, line 155: @GrosPoulet This is unsafe code which is banned by Mozilla, please take a look at my fix in https://github.com/extesy/hoverzoom/commit/98a6d975a24d2aef31209e8d3fdd962c045d6c18

GrosPoulet: @extesy I don't see what's wrong but i trust Mozilla, they are certainly more experts that i am :smile:
I'll apply your changes to StartPage plug-in & move following functions to hoverzoom.js:
- getJsonObjectFromPath
- getKeysInJsonObject
- getValuesInJsonObject

extesy: Yeah, it makes sense to move these functions to a common code since they are used across multiple plugins.

Issue #283, opened 2017-08-02
Images hosted on i.reddit.com not added to history if they are embedded through non-direct link


ragnarkarlsson:
## Description of the problem
When viewing images on https://www.reddit.com/ those which are hosted on the subdomain i.reddit.com but the submitted link is not direct (ie is the featured image of a thread) they are not added to browser history. Yesterday's patch is installed and I'm seeing all imgur links update browser history.

All combos of settings, whitelisting and restarting the browser do not resolve the issue.

## Url of the page where the problem occurs (not the image url)
https://www.reddit.com/r/all

## Steps to reproduce the problem (include image url here)
1. Visit Reddit https://www.reddit.com/r/all
2. Hover over a link to a picture hosted on subdomain i.reddit.com eg https://www.reddit.com/r/aww/comments/6r173k/this_little_guy_waiting_for_his_new_family_to/
3. ???
4. Image displayed, browser history not updated

## OS version, extension version
Win10 (fully patched)
Chrome Version 60.0.3112.78 (Official Build) (64-bit)

extesy:
@ragnarkarlsson Are you sure you are using version `1.0.121` and your option (in "advanced" options tab) is enabled? Because I've just tried to reproduce this and the link does get added to history.

ahoibakk:
I'm having the same issue as @ragnarkarlsson, but it's i.redd.it, not i.reddit.com. It's working for other sites like i.imgur.com though.
Video here: https://img.m42.no/be54b355-45ce-4c8c-917a-e9a5960ee280.mp4
Perhaps it's related to some link trickery reddit is using? HoverZoom is able to magnify the image, but the link is actually to the post, not the image..?

ragnarkarlsson:
Can confirm that I am using 1.0.121 and that the tickbox for _Add viewed pictures to the browser's history_ is enabled.

![image](https://user-images.githubusercontent.com/3290527/28894225-0ede9abe-77cc-11e7-9736-992c41f3017e.png)

Also @ahoibakk has correctly pointed out that it is indeed i.redd.it domain and not i.reddit.com

I'm fairly confident that this was working previously, but couldn't give you anything definitive in terms of timeframe I'm afraid.

extesy:
Hm... I used https://www.reddit.com/domain/i.redd.it/ list and all links were adding into history for me just fine. Maybe some other options and/or plugins interfere with it?

ragnarkarlsson:
I've just tried that URL, result is still the same:

[...]

PR #873, opened 2022-02-13
Fix for "Add to history" functionality

GrosPoulet:


js/hoverzoom.js
```diff
@@ -1296,10 +1296,7 @@ var hoverZoom = {
             setTimeout(posImg, options.showWhileLoading ? 0 : 10);
 
             if (options.addToHistory && !chrome.extension.inIncognitoContext) {
-                var url = hz.currentLink.attr('href') || imgDetails.url;
-                if (url.startsWith('/') && url.indexOf('http') < 0)
```
extesy, line 1300: @GrosPoulet This was done to fix #283 so can you please make sure you didn't revert that fix?

GrosPoulet: @extesy i checked on Reddit, it works fine (because imgDetails.url is now correctly handled).

PR #877, opened 2022-02-18
Rework of Action keys

GrosPoulet:
2 keys added:
 - "Toggle HZ+": quick way to enable/disble HZ+ without using Options page (default key: E)
 - "Close image": close image (or video) being zoomed, instead of just hiding it temporarily (default key: ESC)

Some rewording to make these settings less confusing (i hope)

action_keys

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -507,12 +507,28 @@
         "message": "If a key is set, Hover Zoom+ will be active only when this key is held down.",
         "description": "[options] Action key description"
     },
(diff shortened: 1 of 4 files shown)


extesy:
@GrosPoulet Does `E` disable hoverzoom even after the page reload? I'm concerned that people would accidentally press this key and wouldn't know how to unbreak it (enable it back). It would be safer to leave it unassigned.

GrosPoulet:
@extesy Page reload enable HZ+ back.

PR #887, opened 2022-02-27
New popup for action keys

GrosPoulet:
Some HZ+ users are not aware of existence of action keys and miss out many functionalities (lock, save to disk...).
That's probably because action keys setting is hard to reach, users need to:
- click "Extension menu" puzzle icon
- click "More actions" 3-dots icon
- select "Options" in menu
- select "Action keys" in Options

So i replaced the old popup by a new one for action keys reachable in 2 clicks:
- click "Extension menu" puzzle icon:
popup1

- click "Hover Zoom+":
popup2

Note: classic "Action keys" settings in Options are unchanged.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -59,6 +59,14 @@
         "message": "Options saved. Some settings might need a page refresh.",
         "description": "[options] Notification after saving options"
     },
+    "optCancel": {
+  	"message": "Modification(s) cancelled.",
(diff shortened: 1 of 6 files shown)


extesy:
The old popup also allowed disabling HZ+ per site: 
image

Is it still possible with the new popup?

GrosPoulet:
No, but i can add it to new popup.

extesy:
Yes, please restore per-site disable (second checkbox). I don't think the rest was very useful.

PR #899, opened 2022-03-27
Manifest: Bing plug-in must run in all frames

GrosPoulet:
@extesy you removed **"all_frames": true** directives from manifest for all plug-ins.
This breaks Bing plug-in, so i put it back. Hopefully it does not break any other plug-in...
BTW, why not allowing all plug-ins to run in all frames?

### DIFF
--- a/_locales/ko/messages.json
+++ b/_locales/ko/messages.json
@@ -44,11 +44,11 @@
         "description": "[options] Page title"
     },
     "optButtonDisableAllPlugins": {
-        "message": "Disable all",
+        "message": "모두 사용 안 함",
         "description": "[options] disable all plugins button label"
     },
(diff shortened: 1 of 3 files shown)


extesy:
@GrosPoulet That was my mistake, let me try to explain why I did it. Iframes are very rare in modern internet. In my experience, these days iframes are only used:
1. To show completely separate site in a frame, like bing search.
2. By fans of 90's internet design trends.
3. To inject invisible security exploits or some other nasty stuff.

When I spot-checked some plugins, they worked fine, so I assumed all of this was simply copy/pasted and unnecessary. Some of plugins were indeed copy/pasted, but apparently some (like Bing) actually needed it.

PR #908, opened 2022-04-10
Fix for galleries

GrosPoulet:
This fix improves the way galleries are handled:
- galleries can now contain a mix of images & videos (e.g; Imgur)
- galleries can now contain playlists (m3u8)


GrosPoulet:
@extesy Could you please deliver relase 1.0.190 to Chrome Web Store ?
I'm going to push some quite big changes so i prefer not to have them mixed with previous changes.
It'll make troubleshooting easier in case of regression :fearful:

extesy:
@GrosPoulet Sure, coming up shortly.

PR #925, opened 2022-05-12
FAQ section added

GrosPoulet:
faq


GrosPoulet:
@extesy can you think of other questions/answers that would be useful ?

extesy:
Love it overall! What is `HoverZoom4CherryPics`?

PR #931, opened 2022-05-20
New Advanced option + new plug-in: Dailymotion

GrosPoulet:
There's a new Advanced option: 
option_CORS
By default, this option is **disabled**.
When enabled, HZ+ can now deal with CORS mechanism, as if a specific extension such as SMH (https://chrome.google.com/webstore/detail/simple-modify-headers/gjgiipmpldkpbdfjkgofildhapegmmic), was installed along with HZ+. (Note: SMH is used with Imagus)

The big benefit is that users do not have to install & tweak SMH or similar tool - which is a tedious task - because plug-ins will handle headers rewrite by themselves. 
For now, the only plug-in that uses this feature is **Dailymotion**. As a result, videos from Dailymotion can be zoomed & played from Google & Bing search pages, from Reddit & other sites.

2 optional permissions are added to manifest:
"**webRequest**"
"**webRequestBlocking**"
These permissions are needed to rewrite HTTP(S) request or response headers "on the fly", when they leave or enter the browser.

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -79,6 +79,7 @@ function onMessage(message, sender, callback) {
             break;
         case 'optionsChanged':
             options = message.options;
+            manageHeadersRewrite();
             break;
         case 'saveOptions':
             localStorage.options = JSON.stringify(message.options);
@@ -119,6 +120,9 @@ function onMessage(message, sender, callback) {
         case 'updateViewWindow':
             chrome.windows.getCurrent(window => { chrome.windows.update(window.id, { width:message.updateData.width, height:message.updateData.height, top:message.updateData.top, left:message.updateData.left }) });
             break;
+        case 'storeHeaderSettings':
+            storeHeaderSettings(message);
+            break;
     }
 }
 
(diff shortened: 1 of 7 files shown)


js/background.js
```diff
@@ -164,4 +182,133 @@ chrome.runtime.onInstalled.addListener((details) => {
...
+            let index = headers.findIndex(rh => rh.name.toLowerCase() == h.name.toLowerCase());
+            if (index != -1) headers[index] = { 'name':h.name, 'value':h.value };
+        }
+        if (h.typeOfUpdate == 'add') {
+            let index = headers.findIndex(rh => rh.name.toLowerCase() == h.name.toLowerCase());
+            if (index != -1) headers[index] = { 'name':h.name, 'value':h.value };
+            else headers.push( { 'name':h.name, 'value':h.value } );
+        }
+    })
+    console.log(headers);
```
extesy, line 279: Can you please change all `console.log` into `cLog` to avoid spamming the console for end users?

extesy:
@GrosPoulet This is a very cool feature! Chrome web store requires explanations for each permission and Mozilla's review process is very paranoid so hopefully this doesn't cause any problems.

js/background.js
GrosPoulet: Oups, done in #932

GrosPoulet:
@extesy yes, these permissions are optional, but it may raise some concern, we'll see.
Anyway, there are many extensions dedicated to HTTP(S) headers rewrite (such as SMH) on Chrome Web Store, so there is some hope !

PR #942, opened 2022-06-03
New warning message when page is not focused

GrosPoulet:
Sometimes action keys seem to be broken, this usually happens when page has lost focus.
A warning message will ask user to click anywhere on page so it has focus and action keys can work.
This message is displayed only once (until page is reloaded), so users that do not use action keys are not bothered.

msg


phhorak:
Is there a way to turn it off?

ablankenship10:
Yes please let us turn this off, I don't need any feature other than enlarging images when I hover them, sometimes when im tabbed into a full screen game. This is an extra click on every page load to get this message out of the way.

HeptaEpiThebas:
Also wanted to chime in and reiterate the requests for a setting to disable this.

PR #957, opened 2022-06-16
Fix for plug-in: YouTube_a

GrosPoulet:
- select video with max width available (any format)
- play links with timestart (e.g: https://youtu.be/qORYO0atB6g?t=28)

### DIFF
--- a/plugins/youtube_a.js
+++ b/plugins/youtube_a.js
@@ -1,7 +1,7 @@
 ﻿var hoverZoomPlugins = hoverZoomPlugins || [];
 hoverZoomPlugins.push({
     name:'youtube_a',
-    version:'2.0',
+    version:'2.1',
     prepareImgLinks:function (callback) {
         var name = this.name;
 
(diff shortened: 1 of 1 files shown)


plugins/youtube_a.js
```diff
@@ -83,7 +83,7 @@ hoverZoomPlugins.push({
             cLog('href=' + href);
 
             const re1 = /\/watch\?v=([^&]{1,})/;   // sample: https://www.youtube.com/watch?v=NaOiA15Rz5k
-            const re2 = /\/youtu.be\/([^&]{1,})/;  // sample: https://youtu.be/qXlQbj0PgDo
+            const re2 = /\/youtu.be\/([^?]{1,})/;  // sample: https://youtu.be/qXlQbj0PgDo https://youtu.be/qORYO0atB6g?t=28
```
extesy, line 86: @GrosPoulet But you are still ignoring the actual time value. If the link points to the middle of the video, your code will start playing it from the beginning. You need to add `#t=...` to video url.

extesy: Example: https://www.reddit.com/r/videos/comments/vdg6f7/beastie_boys_intergalactic/ - it should start playing from 28s mark, not from the very beginning.

Issue #986, opened 2022-07-28
Add support for : Ekşi Sözlük


tencirin:
**What is the URL of the website**
https://eksisozluk.com/

**Why we should add support for the website?**
It is one of the, if not the biggest website on the Turkish internet, live since 1999. It was like the Urban Dictionary before Urban Dictionary itself. Although it was actually text based, recently they added image support. 

One example: 
https://eksisozluk.com/entry/140581590

Here the hyperlink with text "görsel" is the link to the image. On the page it is shortened as "https://soz.lk/i/3jadcz74" but when you click on it, it goes to "https://eksisozluk.com/img/3jadcz74". It would be nice to use Hoverzoom on this website.

Thanks

extesy:
Fixed in #988

PR #988, opened 2022-07-29
New plug-in for: eksisozluk.com (#986)

GrosPoulet:


GrosPoulet:
@tencirin i added a plug-in for eksisozluk.com, it'll be available in next release of HZ+ (1.0.195).

PR #1041, opened 2022-10-01
Fix for TikTok plug-ins

GrosPoulet:
When enabled, TikTok plug-ins prevented users from posting comments & liking.


extesy:
@GrosPoulet How bad was the bug in practice? Do I need to do a hotfix release?

GrosPoulet:
Bug prevents users from posting comments & liking.

Le sam. 1 oct. 2022 à 19:11, Oleg Anashkin ***@***.***> a
écrit :

PR #1064, opened 2022-12-29
Update tumblr plugin

dubhater:
Fixes hovering over the new types of links:
"https://www.tumblr.com/<blog name>/<post id>"


extesy:
Thank you for the contribution, @dubhater!

PR #1073, opened 2023-01-29
New plug-ins for: Instagram

GrosPoulet:
Rewrite of Instagram plug-ins, you need be logged in for better efficiency. 
Plug-ins handle:
- pictures
- videos
- carousel (it is possible to download distinct audio tracks & subtitles from videos when present)

@extesy 
- i reduced as much as possible API calls because it looks like they trigger temporary ban, could you please check on your side that it's ok ?
- i added a new script (**BigInteger.js**) to compute media id from shortcode. As BigInt is now supported natively by all browsers it should be possible to do without, maybe you know how ?

### DIFF
--- a/js/BigInteger.js
+++ b/js/BigInteger.js
@@ -1,0 +1,1453 @@
+var bigInt = (function (undefined) {
+    "use strict";
+
+    var BASE = 1e7,
+        LOG_BASE = 7,
+        MAX_INT = 9007199254740992,
+        MAX_INT_ARR = smallToArray(MAX_INT),
+        DEFAULT_ALPHABET = "0123456789abcdefghijklmnopqrstuvwxyz";
+
+    var supportsNativeBigInt = typeof BigInt === "function";
+
+    function Integer(v, radix, alphabet, caseSensitive) {
+        if (typeof v === "undefined") return Integer[0];
+        if (typeof radix !== "undefined") return +radix === 10 && !alphabet ? parseValue(v) : parseBase(v, radix, alphabet, caseSensitive);
+        return parseValue(v);
+    }
+
+    function BigInteger(value, sign) {
+        this.value = value;
+        this.sign = sign;
+        this.isSmall = false;
+    }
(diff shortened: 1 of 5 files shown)


extesy:
I checked it and it works fine for me, gallery mode works too. Didn't notice any issues with request rate limiting. However it only works on somebody's profile page but not on the main https://instagram.com feed - is that expected?

Regarding BigInteger.js - seems like 66ac91bf37f3c434347b783bdd8b54069ab13f4f worked. I replaced `bigInt` with `BigInt`. Did I miss anything?

GrosPoulet:
[https://www.instagram.com/feed/](https://www.instagram.com/feed/) works fine for me.
Replacing `bigInt` with `BigInt` results in a syntax error, e.g: 

> Cannot convert 2<38>not<43><39><47>omq to a BigInt

Check the console.

extesy:
https://www.instagram.com/feed/ view is different from https://www.instagram.com. The former works, the latter doesn't.

Just checked the console, don't see any bigint errors. Are you using the latest master? Do you have a repro link?

extesy:
Ah, it's only available from certain pages like https://www.instagram.com/p/ClpZbUIpNvx/. Let me see if I can fix it.

extesy:
Ok, figured it out: 6ab211f756cc29a0acd51c07fd0de266ed67c215. Should work now.

GrosPoulet:
I do not see anything to zoom on [https://www.instagram.com/.](url)
IG only displays some suggested profiles ('suggestions for you'), but profile pictures can be zoomed (a little)

extesy:
If you follow a few profiles then https://www.instagram.com shows a feed (not to be confused with https://www.instagram.com/feed/ which is just somebody's profile called "feed") of recent posts followed by suggested posts:

image
image

These cards are larger than small squares from the profile view but they still can be zoomed. Exactly the same posts from the profile view are zoomable. As far as I know, https://www.instagram.com is the only real feed (a mix of recent posts from the profiles you follow) that Instagram has when using web UI.

PR #1085, opened 2023-02-12
Add a link to the list of plugins in the readme

Jipok:
The list of supported sites is not easy to find. This is obvious to you as the author, but I needed to go from the addons page to the repository, read the readme, enter a query into the search, and only on the second page of results I understand that each supported site is a separate file in the plugins folder. It took some time. And not every addon user will be able to find it. But I think a lot of people will. 

This PR is just a simple example of how you can help the user find the list. But this is a bad solution to the problem. I would recommend that you provide a link to such information on the addon page. You can also add a list opening from an already installed addon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Hover Zoom+
 ===========
 
-Zoom images/videos on all your favorite websites (Facebook, Amazon, etc). Hover your mouse over any image on the supported websites and the extension will automatically enlarge the image to its full size, making sure that it still fits into the browser window.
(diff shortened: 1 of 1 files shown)


GrosPoulet:
Thanks for the suggestion, @Jipok 
Remark: you can find the list of plug-ins in Options -> Plug-ins :

image

Sometimes names of plug-ins do not indicate clearly which sites are supported, so i'll add logos + tooltips :

image

Jipok:
Where?
[Запись экрана от 2023-02-12 19-52-37.webm](https://user-images.githubusercontent.com/25588359/218318365-b1d4523d-97d0-4af8-b093-d34de454a0f3.webm)

GrosPoulet:
Looks like Options page is only available for Chrome & Edge :confused:

Issue #1069, opened 2023-01-21
Add support for : https://www.temu.com/


ijustliketoread:
**What is the URL of the website**
https://www.temu.com/

**Why we should add support for the website?**
It makes browsing easier! 🙂

Thank you!

GrosPoulet:
Hello @ijustliketoread 
I added a plug-in for temu.com, it'll be available in next release of HZ+ (1.0.203).

PR #1086, opened 2023-02-12
New plug-in for: temu.com (#1069)

GrosPoulet:
This plug-in zooms images & videos provided by users.

### DIFF
--- a/plugins/temu.js
+++ b/plugins/temu.js
@@ -1,0 +1,125 @@
+﻿var hoverZoomPlugins = hoverZoomPlugins || [];
+hoverZoomPlugins.push({
+    name:'temu',
+    version:'0.1',
+    prepareImgLinks:function (callback) {
+
+        var rawData = null;
+
+        var res = [];
+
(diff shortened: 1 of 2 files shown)


plugins/temu.js
```diff
@@ -0,0 +1,125 @@
...
+                    this.addEventListener('load', function() {
+                        try {
+                            let data = this.responseText;
+                            // store relevant data as plain text in sessionStorage for later usage by plug-in
+                            if (data.indexOf('cover_image_url') != -1) {
+                                 // update sessionStorage
+                                let storedHookedDataJson = sessionStorage.getItem('HZOpenData') || '[]';
+                                let storedHookedData = JSON.parse(storedHookedDataJson);
+                                // make some room by removing oldest data
+                                while (storedHookedDataJson.length > 1000000) {
```
extesy, line 42: Why 1000000 in particular? Is there any benefit of strong more than a 100 or even a 1000 responses?

GrosPoulet: Each response is approx. 10k, so we can store around 100 responses before starting deleting oldest responses.

PR #1109, opened 2023-03-19
New plug-in for: Kick.com

GrosPoulet:


GrosPoulet:
@extesy GitGuardian was triggered for no reason imo

extesy:
I think it got spooked by the `token=` query parameter in one of the comments (line 72).

Issue #1127, opened 2023-04-11
The same image gets saved multiple times


desperte:
When hovering an image and pressing the "S" hotkey it downloads the image more than once, usually about 3 or 4 times, sometimes up to 10 times.
I tried changing the hotkey and still the same problem occurs. 

To Reproduce
Try to save an image with the "S" hotkey
Happens on Twitter and Yandex

On the latest version of Chrome and the latest version of HoverZoom

SirGer:
Hover Zoom+ Version 1.0.204  / Chrome (version noted below)

Windows 10 Home 21H2 10.0.19044.2728 / Windows Feature Experience Pack 120.2212.4190.0
HP ENVY x360 m6 Convertible / Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz   2.90 GHz

All of a sudden the Save action is downloading **3 copies** of an image instead of just one; no known updates prior to the onset of this problem except enable/disable another extension as noted below *

Tried updating Chrome to Version 112.0.5615.50 (Official Build) (64-bit) - no joy.

**Removed and reinstalled HZ+ from Chrome Webstore, default settings.** 
The first time I use HZ+ Save (on a thumbnail in a DuckDuckGo image search result page) a pop-up (that does not go away) says HZ+ is requesting additional permissions - allow it to manage downloads? 
**When I click 'Allow' a _single_ download happens. _But from then on any Save is triple_.**

Help? Thx!
* The problem could be associated with having enabled and then disabled an extension called "Always Clear Downloads 2" which was already installed.in Chrome but disabled. I have completely Removed this extension now but the above problem persists.

extesy:
@GrosPoulet Another hotfix needed?

GrosPoulet:
Yep, i fear so. I'll fix it tomorrow.

TS6ix2GaH3q679h5:
Same behavior happening for me.
Version 1.0.204 (last update: Mon, Apr 10, 2023)

mcdoncod:
Same for me

miklosUX:
Same sheite for me 🤬

Is0kaze:
The same problem persists with me... reinstalling the extension or resetting the permissions doesn't solve anything :(

extesy:
@All - the fix is being actively worked on, please wait for updates.

extesy:
This should be fixed now, please try version 1.0.205. Thank you @GrosPoulet for the fix.

SirGer:
[...]

PR #1128, opened 2023-04-14
Fix for multiple downloads of same file (#1127)

GrosPoulet:
@extesy can you test this fix before release please? Download permission is now only checked when setting options.


extesy:
@GrosPoulet Still not fully fixed. I've tried on reddit and intermittently, after saving the same, about 10-15 seconds later, the second "save file" window appears for the same file.

PR #1129, opened 2023-04-15
Fix for multiple downloads when saving a file (#1127)

GrosPoulet:
Sometimes, for some unknown reason, callback for download retry is called even if download succeded.
=> protection added against "parasite" callback calls

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -82,7 +82,7 @@ function onMessage(message, sender, callback) {
                     chrome.downloads.onChanged.removeListener(onChanged);
                     // call callback only if download failed
                     if (delta.state.current !== 'complete') {
-                        callback();
+                        callback(true);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


extesy:
@GrosPoulet Still not completely fixed :(
Repro steps:
1. Go to https://www.reddit.com/r/CCW/comments/12na9ra/always_good_to_see_these_signs_when_we_go_out/
2. Press save image key
3. Cancel the save dialog (don't actually save the image)
4. A few seconds later the save dialog re-appears by itself

GrosPoulet:
@extesy what is the save dialog you are talking about ? The only place i have a save dialog displayed is from Options page:

image

- If i click "Allow" + save  then i can download images (from Reddit or elsewhere) by hovering + pressing hot key (S)
- If i click "Deny" then i can not download images (from Reddit or elsewhere)

In both cases i do not see this dialog anywhere else.

Tested on Chrome version 112.0.5615.121 (Official Build) (64-bit)

extesy:
I mean a file save dialog that Windows shows when downloading the image. I have Chrome set to always ask for directory instead of quietly saving the file.

PR #1139, opened 2023-05-06
Callback added to handle download failure

GrosPoulet:
@extesy callback seems to work fine for me (no duplicate downloads triggered)


extesy:
@GrosPoulet Doesn't work for me in pixiv.net - it now attempts to download html (and shows "save as" dialog) but ends up downloading nothing at all.

GrosPoulet:
@extesy you probably have to re-open the browser after reinstalling extension.
On Pixiv.net, i also have an attempt to download HTML (it fails), but just after HZ+ downloads the image without issue.
Anyway, i re-added the Pixiv.net exception (#1142) so we do not have the HTML download error.

Issue #1174, opened 2023-07-13
kick.com still producing 404 bad request error with hz+ active, after version update meant to fix this issue


eitaniel:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem

greetings

i'm writing to confirm that the fixes you stated you'd perform for hz+ on kick.com seem to not have been implemented. I am still getting 404 bad request error page upon activating hz+ and attempting to surf kick.com. it's almost instant fail. 

did you implement the fixes you suggested would be done according to this issues thread? I've had the addon deactivated as recommended since early June, with the hope of it being fixed as stated in this thread with the next version update (1.0.207), so it's confusing why this seems not to be the case. 

I raised this issue two months ago in this post: https://github.com/extesy/hoverzoom/issues/1156

## Url of the page where the problem occurs (not the image url)

kick.com

## Steps to reproduce the problem (include image url here)
1. activate hover zoom plus, with no site exceptions (ie. normal function to be active on all sites online)
2. visit kick.com, and start surfing normally, therein. 
3. site will produce 404 bad request and eventually not load the site on refresh

## OS version, extension version

windows 10 pro
hz+ v. 1.0.207

![kickstreaming brave browser 400 bad request](https://github.com/extesy/hoverzoom/assets/24715620/1c717ce6-e37b-4fef-bb0d-e61c04516c82)

GrosPoulet:
@eitaniel i did not manage to reproduce the issue. Please remove & add back HZ+ then try again.
BTW, i fixed a bug with clips (#1178), not related with the issue you reported.

eitaniel:
> @eitaniel i did not manage to reproduce the issue. Please remove & add back HZ+ then try again. BTW, i fixed a bug with clips (#1178), not related with the issue you reported.

greetings guys

i just tried using hz+ on opera browser now, freshly installed browser that i'd never used, with freshly installed hz+ extension/addon from chrome webstore. 

went to kick.com with hz+ active (without placing kick.com as an exception site, ie fully operational on kick.com). and the moment i clicked the kick icon on top left of screen, to essentially refresh the page, instantly opera gave me a 'bad request' blank page. 

the issue still exists. I personally never mentioned something about clips, my main issue was that the website itself seems to break completely, with hz+ active/on. I thought i'd made this clear in my original issue report, and that the issue was understood and had been fixed as stated for this current (new) version. 

[...]

PR #1191, opened 2023-08-06
Fix for Custom Headers

GrosPoulet:


extesy:
@GrosPoulet Would this fix the kick.com problem from #1174?

GrosPoulet:
@extesy no it is not related to kick.com issue

PR #1196, opened 2023-08-09
fix redgifs cors

yadue:


plugins/redgifs.js
```diff
@@ -25,7 +25,22 @@ hoverZoomPlugins.push({
...
+                    value: `Bearer ${tempToken}`
+                }]
+            }, (response) => {
+                var data;
+
+                try {
+                    data = JSON.parse(response);
+                } catch (e) {
+                    return;
+                }
```
extesy, line 42: Why swallow exceptions? When this approach eventually breaks, it will be easier to debug if it logs to console.

PR #1200, opened 2023-08-10
fix cors issue

yadue:
I had to redo the commit, creating new pull request. 

About the try {} catch {}. Literally every single JSON.parse is made like that in that repository.

### DIFF
--- a/plugins/redgifs.js
+++ b/plugins/redgifs.js
@@ -15,17 +15,26 @@ hoverZoomPlugins.push({
             }
         });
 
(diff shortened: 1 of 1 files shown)


extesy:
@yadue I've checked your change on https://www.reddit.com/search/?q=site%3Aredgifs.com&sort=new&t=all and it works for videos (`www.redgifs.com` links) but does not work for images (`i.redgifs.com` links). Would you be able to make it work with images as well?

yadue:
@extesy added

PR #1220, opened 2023-08-21
add support for redgifs links including http

yadue:
currently selector does not support http link, this should work for all giphy links

### DIFF
--- a/plugins/redgifs.js
+++ b/plugins/redgifs.js
@@ -15,7 +15,7 @@ hoverZoomPlugins.push({
             }
         });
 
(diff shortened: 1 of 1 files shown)


extesy:
@yadue Unfortunately that's not possible because browsers don't allow connections from https to http links anymore.

extesy:
@yadue Do you have example links to verify this? Usually browsers don't allow https->http connections.

yadue:
@extesy its not opening those http links, see the whole file. it only detects all links with http and https and extracts giphy id to use it for api call. at this stage I don’t have any example at this point but I see Reddit randomly using http or https links.

PR #1233, opened 2023-09-24
Save playlists (M3U8) to audio & video files

GrosPoulet:
A playlist is saved to 3 distinct files:
- playlist itself (M3U8)
- audio (MP3)
- video (MP4)

TODO : mux audio & video into one single file.
This could be done with ffmpeg.js, but this lib is pretty big :worried:
@extesy any better idea ?


extesy:
Muxing would be an overkill, let's not add too much bloat to the extension. I think it's fine to keep files separate. Maybe add FAQ entry on how to merge them, with a sample ffmpeg command line?

Issue #1238, opened 2023-10-01
Not working on Flickr 


pwmcmaho:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem
Nothing is happening when viewing images on the explore section of the Flickr website:

## Url of the page where the problem occurs (not the image url)
https://www.flickr.com/explore

## Steps to reproduce the problem (include image url here)
1.  Go to the website: https://www.flickr.com/explore
2.  Hover over an image
3.  Notice that hoverzoom does not appear to recognize or react to the image

## OS version, extension version
MacOS 11.17 (BigSur)
Firefox 117.0.1

wlm2048:
I can add the following details, this is the error occurring in the logs:
VM62 flickr_a.js:106 [HoverZoom] Flickr API call failed. Photo ID: 53257554450. Error #100: Invalid API Key (Key has expired)

Win 11/Chrome 117/HZ 1.0.210

GrosPoulet:
@wlm2048 I fixed plug-ins for Flickr, they'll be available in next release of HZ+ (1.0.211).

PR #1248, opened 2023-10-20
Fix for Flickr plug-ins (#1238)

GrosPoulet:


GrosPoulet:
API key updated (courtesy of Imagus) + others improvements.

PR #1251, opened 2023-10-25
README.md

RS-labhub:
Fixed grammatical errors and typos

PR #1252, opened 2023-10-25
Update README.md

suhit123:
improved readme


extesy:
#1251 was the first to suggest the same changes

PR #1253, opened 2023-10-26
Fix Typo errors in CHANGELOG.md

FarukhS52:


extesy:
@FarukhS52 Changelog is auto-generated from the list of commits, so unfortunately all of these changes will be overwritten the next time changelog is updated.

PR #1256, opened 2023-11-01
New plug-in for: Google Drive + fix for Google Photos + old bug fix

GrosPoulet:
Google Drive plug-in zooms images & also non-images documents, such as XLS files, PowerPoint,...

image

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1681,7 +1681,7 @@ var hoverZoom = {
                     imgThumb = hz.currentLink.find('[style]');
                     if (imgThumb.length > 0) {
                         imgThumb = $(imgThumb[0]);
-                        lowResSrc = hz.getThumbUrl(imgThumb);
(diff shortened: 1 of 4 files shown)


js/hoverzoom.js
```diff
@@ -1681,7 +1681,7 @@ var hoverZoom = {
                     imgThumb = hz.currentLink.find('[style]');
                     if (imgThumb.length > 0) {
                         imgThumb = $(imgThumb[0]);
-                        lowResSrc = hz.getThumbUrl(imgThumb);
+                        lowResSrc = hz.getThumbUrl(imgThumb[0]);
```
extesy, line 1684: @GrosPoulet Does this work correctly? Line 1683 takes the first element of the list. But then line 1684 does it again? Was it a nested list?

GrosPoulet: @extesy function `getThumbUrl(arg)` takes a DOM element as argument, not a JQuery object.

PR #1302, opened 2024-01-31
feat: store redgifs token to avoid hitting rate limit

yadue:
Redgifs started to throw 429 Rate limit when requesting to often to temporary token. This PR stores it in variable and reuses it.


extesy:
@yadue I don't think this will make any difference. The token is only requested once per `prepareImgLinks` call which only happens on page load.

yadue:
@extesy I wouldn't work on this fix if it wasn't required. New reddit has "infinite scroll" and with each next page load it triggers prepareImgLinks to run. It requires just couple of runs for redgifs to throw 429.

Issue #1283, opened 2024-01-07
Chrome hangs with Hover Zoom+ enabled


sdr0x07b6:
I am currently developing a web application. It has the ability to add avatar photos by dragging and dropping them.
When I add a photo to that page with Hover Zoom+ enabled, a few seconds after adding it, the CPU usage goes to 100%, the page will not accept any input or operation, and I have no choice but to crash it in Task Manager.

If I disable Hover Zoom+, reload, and then do the same, the problem does not occur. If I re-enable it and reload, the problem still occurs. Reproducibility is 100%.
It does not seem to hang up immediately after adding a photo, but after about 5-10 seconds after adding the photo.

The application under development has been in development for some time. This phenomenon did not occur at that time.
Development stopped for several months, and it occurred when we resumed it today. So I think it could be caused by the relatively recent release of Hover Zoom+.

I don't know how to look at the Performance tab in Developer Tools as I don't use it very often.
Is there a hint in the display `"Function prepareImgLinks @ hoverzoom.js:2076:33"`?
![hoverzoom+](https://github.com/extesy/hoverzoom/assets/1102138/dc9bbfa8-7232-4e2c-a4b3-33abefed5a73)

- Hover Zoom+ 1.0.212
- Chrome for macOS 120.0.6099.199 arm64

extesy:
`prepareImgLinks` is at the high level of a stack trace. Can you find what function at the bottom of the stack is causing this?

sdr0x07b6:
@extesy Hello. I have not been able to make time for you. Sorry for the wait.

I see a lot of jQuery in the stack trace.
![Screenshot - 2024-01-13 16 01 55](https://github.com/extesy/hoverzoom/assets/1102138/cc6cb91b-bb22-4b83-8661-491b901a0caa)

I am not quite sure how to use the Performance tab, so if there is more that needs to be done please direct me.

The version of Chrome was recently upgraded.
- Chrome for macOS 120.0.6099.216 arm64

extesy:
@sdr0x07b6 you can see that each frame points to the source file and a line number, specifically `google_a.js:80` and `google_a_js:71`. Looking at these lines, the jquery filters definitely look very broad, so there's opportunity for optimization there. In the meantime you can disable that plugin in extension settings.

yadue:
```
const filter2 = /(?=.*\.googleusercontent\.com\/)(?!.*=)/;
const regex2 = /(.*)\/.*\//;
[...]

PR #1303, opened 2024-01-31
feat: temporarily disable regex causing huge performance issues

yadue:
This PR disabled changes introduced by @GrosPoulet with https://github.com/extesy/hoverzoom/pull/1256, only the lines causing performance issues. Note here that I won't work on fixing regex itself.

Fixes: https://github.com/extesy/hoverzoom/issues/1283


extesy:
@yadue can you please check if the previews on Google Images search page still works?

yadue:
@extesy for me it works but, I clearly disabled some case which I'm not going to investigate nor fix. In this PR I'm disabling horrible performance issues with some websites. For me it literally breaks browser tab.

extesy:
@GrosPoulet I'm going to merge this PR and if we discover that it broke a preview somewhere then let's fix it in a less performance-impacting way.

Issue #1301, opened 2024-01-30
Add Alt key as an option for "Activate Hover Zoom"


Goosegit11:
**Is your feature request related to a problem? Please describe.**
Currently there's no Alt key option in "Activate Hover Zoom key" setting:
![image](https://github.com/extesy/hoverzoom/assets/89806596/56d3d0bf-c677-4283-9787-cc2e21df9904)

**Describe the solution you'd like**
Add Alt key as an option.

**Describe alternatives you've considered**
Imagus mod can do that:

![image](https://github.com/extesy/hoverzoom/assets/89806596/b5fa2f59-03b8-40c6-b321-f3e91d25dba7)

**Additional context**
Some people might think this is inconvenient because the menu bar appears every time you press Alt, but you can actually disable this behavior by setting `ui.key.menuAccessKeyFocuses` to false. (That's for Firefox)

PR #1304, opened 2024-01-31
Adding Alt key as an option for "Activate HoverZoom+".

nymver:
Solves issue #1301


extesy:
@nymver Does it work both in Chrome and Firefox?

nymver:
> @nymver Does it work both in Chrome and Firefox?

Yeah, I tested it.

Goosegit11:
Yep, I've been using it for a few days now and it works just great.

Issue #1300, opened 2024-01-30
Allow specyfying max size for the preview


Goosegit11:
**Is your feature request related to a problem? Please describe.**
I always get frustrated when the preview is too big. I want to control its size from the addon settings.

**Describe alternatives you've considered**.
The Imagus mod has this feature, it allows you to specify a maximum width and height:

![image](https://github.com/extesy/hoverzoom/assets/89806596/403aceb7-61b4-43ff-ad54-06f5deffcb04)

**Additional context**
I tried to do this via the enlarge zoom factor option, with set it to 0.5, but it doesn't accept decimal numbers.

**Describe the solution you'd like**.

I can see these solutions, but you can come up with your own way as well:

1. Do it the Imagus mod way, introducing max width and height.
2. Allow set factor to decimals and numbers below 1.
3. Introduce some kind of percentage. Like, set it to 75% and preview will be 75% of the original size.

PR #1309, opened 2024-02-07
Add max width to zoomed pictures.

nymver:
Solves issue #1300


nymver:
Again, my friend @Goosegit11 tested it on Firefox and I tested it on Chromium. Everything works.

extesy:
@nymver Thank you for the pull request! The original request is asking for the maximum _size_, not just the _width_. There could be portrait-mode images what your PR will not limit. To actually address #1300 you need to 1) rename `maxWidth` to `maxSize`, 2) limit both width and height.

PR #1317, opened 2024-02-29
Fix mediawiki plugin thumbnail selector

reohjs:
Use a less precise attribute selector since mediawiki thumbnail URLs no longer use the full URL

E.g. https://runescape.wiki/


extesy:
I'm concerned that the new selector `img[src*="thumb/"]` is too generic and will have false positive matches of various non-wiki websites. Is there anything else about mediawiki urls that would help make the selector narrower?

PR #1324, opened 2024-04-01
Fix for plug-in: drupal_a

GrosPoulet:


GrosPoulet:
@extesy Happy Easter, it's release time :tada:

extesy:
@GrosPoulet Coming right up!

By the way, we really need to do something about Manifest V3 and soon. Otherwise:
> More specifically, if you have not migrated to Manifest V3 by June 2024, your Manifest V2 extension may begin to be disabled on Chrome browsers and the relevant item will lose its Featured badge in the Chrome Web Store if it currently has one.

GrosPoulet:
Yes, do you know how to perform this migration to Manifest V3?

extesy:
I don't know, but Google has published a [migration guide](https://developer.chrome.com/docs/extensions/develop/migrate) for that. I think the migration can be done in multiple stages (service workers, api updates, etc) so you are welcome to take a look and maybe start with some simple low-hanging fruit. I suspect (not 100% sure) that V3 is backwards compatible so after all migrations are done then V2->V3 switch should be a no-op.

PR #1372, opened 2024-07-06
New option: add caption to filename when saving

GrosPoulet:
For instance, an image of Mona Lisa is saved as: `[Mona Lisa by Leonardo da Vinci]image03.jpg`
By default, this option is **disabled**.

This new option is available in Advanced options:

image


GrosPoulet:
@extesy release please ^^

PR #1419, opened 2024-09-24
Fix: LinkedIn profile pictures with 'hiring' or 'open to work' tags not zooming correctly

sohel0707:
> This PR fixes an issue with LinkedIn profile pictures containing tags like 'hiring' or 'open to work' that were not zooming correctly in the HoverZoom extension.

Previously, the code assumed that the 4th image in the LinkedIn API response was always the largest one. However, in cases where profile tags such as 'hiring' or 'open to work' are present, the 4th image is actually the smallest. This PR updates the image selection logic to handle such cases, ensuring that the correct (largest) image is used for zooming, regardless of the presence of these tags.


extesy:
Thank you for the contribution, @sohel0707!

Issue #429, opened 2019-02-02
Not working on Furaffinity


Fayings:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem
Add-on doesn't work on Firefox on the website Furaffinity

## Url of the page where the problem occurs (not the image url)
https://www.furaffinity.net/

## Steps to reproduce the problem (include image url here)
1. Every single image- hovering with the mouse over it does nothing.

The add-on works with the website if you use it on Chrome, but not on Firefox.

PR #1436, opened 2024-10-27
Fix: Function prepareFromDocument() breaking on Firefox

LiliaDoe:
In hoverzoom.js, changes:
```js
doc.open(); 
doc.write(data);
doc.close();
```
to 
```js
doc.body.innerHTML = data;
```

- This fixes it giving the following error on Firefox: `"SecurityError: The operation is insecure."`
- Also updated variable scopes to match the style guide.

This will fix some plugins on Firefox that rely on prepareFromDocument() to function. Issue #429 is fixed with this, and there might be more. I tested it on Chrome and Firefox, and it worked for me.


extesy:
Seems reasonable. Thank you, @LiliaDoe!

Issue #1387, opened 2024-08-07
Hover Zoom+ for Firefox - "Save image" Action Key does not work (Version 1.0.218)


isick1e:
**Describe the bug**
On Version 1.0.218 extension for Firefox, the "Save image" Action Key does not work to save an image that is currently hovered. This functionality works fine on the Chrome extension, however. 

All of the other Action Keys seem to work, aside from "Save image". I have just moved over to Firefox from Chrome, so I don't know how long this function hasn't worked. 

I tried to reinstall the extension fresh with all settings reset using the "Reset" and "Save" buttons to no avail. As seen in the screenshots below, under the extension permissions, only "Access your data for all websites" is enabled by default. 

I went ahead and enabled the three other Optional permissions "Download files and read and modify the browser’s download history", "Access browsing history", and "Access browser tabs". The "Save image" Action Key still doesn't download images. I have also enabled the "Enable Action Keys focus overlay message" setting to ensure the image is focused before trying to save it.

**To Reproduce**
1. Go to Google.com
2. Search "cats"
3. Click on Images
4. Hover over various image results and allow them to enlarge.
5. Press the "Save image" Action Key (default S)
6. Nothing happens, the image does not download.

_In the Screenshots section below, I will show the Console outputs when using the "Save image" Action Key._

**Expected behavior**
Pressing the desired Action Key on a hovered image should download the image to the set download location.

**Screenshots**
**_Screenshots of the default Hoverzoom+ Extension settings on Firefox:_**
![firefox_7r0Nfryllx](https://github.com/user-attachments/assets/69c8ff93-8a12-4a8e-a3f1-2f8fb9043520)

![firefox_ZjzspzdlWh](https://github.com/user-attachments/assets/76d1ef2e-6ffd-41c8-9001-cfe177056b66)

**_Screenshots of my modified Hoverzoom+ Extension settings on Firefox to enable optional features:_**

![firefox_Fuv10tL4rG](https://github.com/user-attachments/assets/c2d36e2f-5bfc-4537-a3ff-75f7d87f405f)

_**Screenshot of the Firefox Console (F12) and the Hover Zoom+ Extension debug console outputs when hovering over the first 3 Google image results and pressing the "Save image" Action Key.**_

![firefox_nBHuxJhcBC](https://github.com/user-attachments/assets/6a67c92d-c920-44e0-ad3f-2c227a2ba10c)

[...]

PR #1437, opened 2024-10-27
Fix: "Save image" Action Key does not work

LiliaDoe:
Fix for #1387
Changes:
```js
chrome.permissions.request({permissions: ['downloads']}, function (granted) {
    cLog('downloadFile granted: ' + granted);
    if (granted) {
        ...
    }
});
```
to
```js
chrome.permissions.contains({permissions: ['downloads']}, (contained) => {
    cLog('downloadFile contains: ' + contained);
    if (contained) {
        ...
    } else {
        chrome.permissions.request({permissions: ['downloads']}, (granted) => {
            cLog('downloadFile granted: ' + granted);
            if (granted) {
                ...
            }
        })
    }
});
```

This fixes images not downloading on Firefox, and still allows chrome to request permissions via a prompt. It is not an ideal fix, but until this bug is resolved on Firefox's end, this is the best fix I found.

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -80,13 +80,27 @@ function downloadFile(url, filename, conflictAction, callback) {
 function onMessage(message, sender, callback) {
     switch (message.action) {
         case 'downloadFileBlob':
-            // direct URL download through Chrome API might be prohibited (e.g: Pixiv)
-            // workaround:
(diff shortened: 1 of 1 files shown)


extesy:
Amazing. Thank you, @LiliaDoe!

Issue #633, opened 2020-11-29
Pixiv.net - Saving with shortcut does not work


Autistic-Unicorn:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem
Saving image when in hover mode is not working, outputs 
`Access to fetch at 'https://i.pximg.net/(link to image there).jpg' from origin 'https://www.pixiv.net' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.`

## Url of the page where the problem occurs (not the image url)
pixiv.net

## Steps to reproduce the problem (include image url here)
1.Hover over image
2.Load the image in Hover zoom mode
3.Try to save image with "S" shortcut

## OS version, extension version (Chrome only, Firefox is not supported)
Windows 10, 1.0.158

Autistic-Unicorn:
It is possible to make saving with shortcut to work with this addon: https://mybrowseraddon.com/access-control-allow-origin.html
Sadly i don't know anything in-deep of that issue so i just hope i can help with giving a good path to fixing the issue.

extesy:
Fixed in bdbc9f4d6c570c2106d739c99f37b6bb72fcd307, will be available in the next release.

extesy:
Still not fixed because of pixiv's referrer check.

extesy:
Fixed in https://github.com/extesy/hoverzoom/commit/130ae06d83ab2ff28cbe579979862cc0666b26c5, will be available in the next release.

PR #1438, opened 2024-10-28
Fix: Pixiv.net - Saving with shortcut does not work

LiliaDoe:
Fix for #633
- Also updated the variable scopes and comments in background.js to match style guide.
- Tested on Firefox and Chrome.

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -274,15 +274,19 @@ chrome.runtime.onInstalled.addListener((details) => {
     }
 })
 
-// - store request's header(s) setting(s) = modification(s) to be applied to request's header(s) just before sending request to server
-//   e.g: add/modify "Origin" header to deal with CORS limitations
-// - store response's header(s) setting(s) = modification(s) to be applied to response's header(s) just after receiving response from server
-//   e.g: add/modify "Access-Control-Allow-Origin" header so browser allows content display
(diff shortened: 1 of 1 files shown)


js/background.js
```diff
@@ -359,26 +365,43 @@ function updateHeaders(headers, settings) {
...
+* - onHeadersReceived
+* also remove headers settings since they are not used anymore
+*/
 function removeWebRequestListeners() {
-    if (chrome.webRequest.onBeforeSendHeaders.hasListeners())
+    if (!chrome.webRequest.onBeforeSendHeaders.hasListener(updateRequestHeaders))
         chrome.webRequest.onBeforeSendHeaders.removeListener(updateRequestHeaders);
-    if (chrome.webRequest.onHeadersReceived.hasListeners())
+    if (!chrome.webRequest.onHeadersReceived.hasListener(updateResponseHeaders))
         chrome.webRequest.onHeadersReceived.removeListener(updateResponseHeaders);
```
extesy, line 405: This doesn't look right: if there is no listener then you are removing them? Probably a copy/paste typo.

LiliaDoe: oh, yeah. It is

extesy:
Thank you!

PR #1439, opened 2024-10-29
Fix: Hovered images flicker on Fur Affinity

LiliaDoe:
I fixed images flickering when you move your mouse around while hovering over one.
- Made prepareFromDocument callback function return a full url
```js
    return img ? img.dataset.fullviewSrc : null;
```
to
```js
    return img ? 'https:' + img.dataset.fullviewSrc : null;
```
- Updated variable scopes


extesy:
Thank you!

Issue #1216, opened 2023-08-18
Option to automatically lock zoomed images to center of the page


icp1994:
**Is your feature request related to a problem? Please describe.**
For my use cases of hoverzoom, I almost always want to lock the hovered images for pan/zoom.

**Describe the solution you'd like**
It would be nice to have an option to auto-lock zoomed images rather than having to use an additional hotkey. Especially since I use `Ctrl` modifier for `Activate HoverZoom+`, using it with another key interferes with browser default hotkeys.

**Describe alternatives you've considered**
I have briefly searched for workarounds to lock without using hotkey.

**Additional context**
N/A

extesy:
@icp1994 You can use right click to do that:
![image](https://github.com/extesy/hoverzoom/assets/65872/357de177-a52f-482b-80e5-075e6e8edf85)

icp1994:
True, that's what I used at first. But it also triggered browser's context menu (open in new tab, open in new window, save as etc) which was a bit annoying often times.

I think even without implementing this feature, if you allow `ALT` as an action key that will solve my problem since I don't have browser shortcuts that would conflict with Alt modifier.

TraceGroot:
I have to agree with this issue.  It's a royal pain to have to hold down the enable key and/or have to separately press the L key in order to get an image to lock in place.

This is a very easy item to implement via a simple configuration setting.  "Automatically Lock when Activating" - On/Off.

This one issue is what keeps me using Imagus instead of HoverZoom+.

extesy:
Fixed by 1d85309a0f2473bbe226cf524dc2e7ab25738e99, will be available in the next release.

PR #1440, opened 2024-10-30
Add 'Automatically lock images' function

LiliaDoe:
Requested option from issue #1216

- Locks image whenever Hover Zoom activates. Added option for it in the options page. Only has English name and tooltip.

- Tested in Firefox and Chrome

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -34,7 +34,7 @@ var factorySettings = {
     ambilightBackgroundOpacity : 0.9,
     disabledPlugins : [],
     centerImages : false,
-    autoLockImages  : false,
+    autoLockImages : false,
     frameBackgroundColor: "#ffffff",
     frameThickness: 4,
     displayImageLoader: false,
--- a/js/common.js
+++ b/js/common.js
@@ -34,6 +34,7 @@ var factorySettings = {
     ambilightBackgroundOpacity : 0.9,
     disabledPlugins : [],
     centerImages : false,
+    autoLockImages  : false,
     frameBackgroundColor: "#ffffff",
     frameThickness: 4,
     displayImageLoader: false,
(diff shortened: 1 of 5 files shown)


js/common.js
```diff
@@ -126,6 +126,7 @@ function loadOptions() {
     options.ambilightBackgroundOpacity = options.hasOwnProperty('ambilightBackgroundOpacity') ? options.ambilightBackgroundOpacity : factorySettings.ambilightBackgroundOpacity;
     options.disabledPlugins = options.hasOwnProperty('disabledPlugins') ? options.disabledPlugins : factorySettings.disabledPlugins;
     options.centerImages = options.hasOwnProperty('centerImages') ? options.centerImages : factorySettings.centerImages;
+    options.autoLockImages = options.hasOwnProperty('autoLockImages') ? options.autoLockImages : factorySettings.autoLockImages;
```
extesy, line 129: Please also add `autoLockImages` to the `factorySettings` constant at the beginning of this file, with the default `false` value.

LiliaDoe:
@extesy This may be better as a new issue, but I didn't want to make one.

 Upon testing the auto lock function again, I did not implement it well. It'll lock, but without respecting the new zoom options. Plus, when locked, pressing the lock key does nothing and scrolling sometimes doesn't zoom. I have been playing with it, but I'm not entirely sure how to fix it and could use help. It seems like a simple feature to add properly, but I haven't figured out how to make it behave as if one just pressed the "Lock Image" key.

extesy:
@LiliaDoe Would it be better to revert this PR until a proper fix is found?

LiliaDoe:
I think the people who use it are probably fine with it behaving as it is, but I would like it to behave like the lock image key ideally

Issue #570, opened 2020-09-02
Activate full zoom not working


drhouse:
**Describe the bug**
The following hotkey in the options menu does not zoom any picture regardless of which key it's mapped.
_"Activate full zoom - When this key is held down, the picture is displayed using all available space. Useful for high resolution pictures only."_

**To Reproduce**
1. Go to https://www.google.com/search?q=lakes
2. Hover over the first lake image
3. Hold hotkey 'Z' down, which is the default mapping for 'Activate full zoom'

**Expected behavior**
Some sort of zoom effect should take place. 

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome x64
 - Version: 85.0.4183.83

GrosPoulet:
Hi Doctor
That's because in some cases HZ+ does not find any bigger version of picture being hovered.
So it may just display the same picture, this is not an issue with key mapping.
In your Google Images example, zoom works on most lake pictures but not all of them.

drhouse:
Thanks for the reply, but in my case the hotkey 'Z' does nothing to any image... 
the standard hover-zoom feature works and shows the 'full-sized image' over the thumbnail, but I thought the 'Z' key was supposed to further expand the 'full-sized image' to use up as much as the browser screen space as available (while being constrained by ratio).

extesy:
@drhouse Do you have "Extend zoomed images below the mouse cursor" option enabled by any chance? The zoom key is basically a temporary on-demand version of that option.

EpochalEngineer:
I also recall the original HoverZoom to work that way: zoom to all available space, even if there is not a larger source version available.

I would much appreciate a feature to do this, as I often browse on a 4K resolution with much of the source material being small in comparison.

drhouse:
> I also recall the original HoverZoom to work that way: zoom to all available space, even if there is not a larger source version available.
> 
> I would much appreciate a feature to do this, as I often browse on a 4K resolution with much of the source material being small in comparison.

for what it's worth, soon after my last post I discovered Imagus, it does everything the way I want it to work and seem to remember Hoverzoom working, highly recommended:
[chrome store - Imagus](https://chrome.google.com/webstore/detail/imagus/immpkjjlgappgfkkfieppnmlhakdmaab)

EpochalEngineer:
I actually switched from Imagus to HoverZoom+ because Imagus has issues with v.reddit.com.

But yes, I definitely prefer the Imagus behavior other than that.

[...]

PR #1442, opened 2024-10-31
Fix: Fullzoom button not filling window

LiliaDoe:
From issue #570

- When pressing the 'Full Zoom' button, the image will expand to window's size + padding.
- Fullzoom might look better if the image could stretch to completely to bottom of window, but this implementation works.
- Tested on Firefox and Chrome


LiliaDoe:
Just realized this might be a duplicate PR

LiliaDoe:
That seems to implementing a different feature, so I'll leave this PR up

PR #1449, opened 2024-11-03
Fix for plug in: Gifbin

LiliaDoe:
- Status code 206 "Invalid URI. Load of media resource failed." shows after loading each gif, but they still load and play as expected.


extesy:
> Status code 206 "Invalid URI. Load of media resource failed." shows after loading each gif

A bit annoying but I guess that's better than not working at all. Thank you!

Issue #1390, opened 2024-08-15
Doesn't work on the reddit.com in Compact View


scotto88:
**ATTENTION!** If you don't follow this template then your issue will be immediately closed.

## Description of the problem

Doesn't work on the reddit.com in Compact View.

## Url of the page where the problem occurs (not the image url)

https://www.reddit.com/ but need to changed to Compact View instead of Card View.

## Steps to reproduce the problem (include image url here)
1. browse to reddit.com
2. Change view to Compact.
3. Try to hover over an image that would usually zoom out.

## OS version, extension version
Windows 11
Extension 1.0.218

khamseen:
Glad someone already reported this, another reason to resent them forcing the latest UI on us...

Issue #458, opened 2019-09-02
imgur gifs (videos) on reddit don't play anymore


Eiion:
## Description of the problem
Gifs on imgur can't be previewed on reddit anymore. I first noticed this about 1-2 weeks ago.

## Url of the page where the problem occurs (not the image url)
imgur.com

## Steps to reproduce the problem (include image url here)
1. Surf reddit
2. hover over imgur link
3. see preview picture but not the playing gif/video

## OS version, extension version (Chrome only, Firefox is not supported)
Windows 10 Pro 64bit, Chrome Version 76.0.3809.132 (Official Build) (64-Bit)

EDIT:
## Additional info:
Attention: The following links might be considered NSFW
Oddly enough it works for some but not for other links on reddit.
While this link works: https://imgur.com/a/dB2kDnx
this one doesn't:        https://imgur.com/a/6GWnjBk
Both are the same format, both apparently are stored as *.mp4.

LiliaDoe:
neither of those links seem to be valid anymore. Can you provide new examples?

extesy:
This is a very old bug report. Most likely it's been already fixed a long time ago. I'm closing it.

PR #1452, opened 2024-11-04
Fix: sh.reddit.com search and compact view not loading

LiliaDoe:
Made sh.reddit search and compact view media load.
- In searches, selecting only the thumbnails allowed it to load, while in normal compact view, selecting the whole faceplate allowed it to load. Not sure why. I want to streamline it, but for now, things work as expected.
- Might allow searches to load media in old.reddit as well in the future

### DIFF
--- a/plugins/reddit.js
+++ b/plugins/reddit.js
@@ -115,9 +115,13 @@ hoverZoomPlugins.push({
     // To load sh.reddit videos
     $('shreddit-player-2').one('mouseover', function () {
       let post = $(this);
-      let packagedMedia = post.attr('packaged-media-json');
-      let link = 'https://' + packagedMedia.match(/"source":{"url":".*(packaged-media.redd.it\/.*)","dimensions"/)[1];
-      hoverZoom.prepareLink(post, link);
+      let src = post.attr('src')
+      let packagedMedia = post.attr('packaged-media-json')
+
+      if (packagedMedia) {
+        src = 'https://' + packagedMedia.match(/"source":{"url":".*(packaged-media.redd.it\/.*)","dimensions"/)[1];
+      }
+      hoverZoom.prepareLink(post, src);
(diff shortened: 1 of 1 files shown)


LiliaDoe:
I'll work on reducing the duplicatiton later

extesy:
When you are going to test this, please make sure that old.reddit.com still works fine. Old reddit is the best reddit.

LiliaDoe:
Do I need to reduce the duplication? I can work on it, but I see that was there from previous commits. I'm a little afraid to change any of it in case some part of old.reddit with a gallery stops working, to be honest.

LiliaDoe:
I've tried reducing the duplicate lines, and i think I got it as low as I can without changing previous code too much. There were more errors with my changes than I expected, but after looking over it for a while, everything is working. The places I tested on old.reddit.com and sh.reddit.com are: galleries, images, and videos, in subreddits and home feed. The only thing not working is searches on old.reddit.com, something that wasn't working previously.

LiliaDoe:
I looked over the whole file again, and made sure the old code was how it was when I found it, with updated variable scopes. My only concern is new reddit tends to run slower than old reddit, but I think that's just the nature of the websites rather than HZ+

extesy:
Seems to work for me. Thank you!

LiliaDoe:
Also, this fixes #1390 and possibly #458

LiliaDoe:
If users report issues with sh.reddit.com searches not loading when mouse hasn't moved, one could resolve it by changing:
```js
    $('faceplate-tracker[data-faceplate-tracking-context*="post_thumbnail"]').each(function () {
```
to
```js
    $('faceplate-tracker[data-faceplate-tracking-context*="post_thumbnail"]').one('mouseover', function () {
```
Using .each() just makes images load faster

PR #1454, opened 2024-11-06
Fix: GitHub not loading private user images

LiliaDoe:
Images on github with a src containing "private-user-images" wasn't loading before. This fixes that. I discovered the issue when looking at issues on the hz+ repository.


extesy:
Oh, nice!

PR #1455, opened 2024-11-06
Fix: Zoomed images's horizontal position wrong on Fur Affinity on 2k monitors

LiliaDoe:
Variable 'body100pct' in hoverzoom.js is not picking up Fur Affinity using max-width which sets hzWindow position wrong when innerWidth is greater than max-width. This fix makes body100pct check if max-width is changed in the body's css, returning false if it is.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -485,7 +485,7 @@ var hoverZoom = {
                 }
 
                 // width adjustment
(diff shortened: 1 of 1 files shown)


LiliaDoe:
Not sure why its showing past changes in the changed file. Everything is back up to date though

PR #1456, opened 2024-11-07
Fix plugin eksisozluk not working on chrome

LiliaDoe:
It seems the site requires a referrer header when making an ajax request, which isn't being sent with chrome. I made a workaround by creating a link from the post's date and the link's href. There's probably a better method for this though, e.g. sending a referrer header with the ajax request

### DIFF
--- a/plugins/eksisozluk.js
+++ b/plugins/eksisozluk.js
@@ -2,26 +2,21 @@
 hoverZoomPlugins.push( {
     name: 'eksisozluk',
     version: '1.0',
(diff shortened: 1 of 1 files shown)


LiliaDoe:
I'm not great at debugging xml requests, so if there's a way to remedy the referrer heading not sending on chrome, I think that'd be ideal

extesy:
I'm not great at that either. Anything that works and is not breaking anything else is good enough for me. Thank you!

Issue #804, opened 2021-10-04
Losing settings on Chrome cache removal?


miklosUX:
It seems I lose the custom settings, like whitelisted sites, keys to control functionalities... every time I delete the browser cache (nothing else, i.e. cookies) in Chrome, macOS (latest both).

A giant PITA to have to go back in and set everything back up. (Someone suggested the ability to export/import settings btw which I agree with, it would really help!)

**Desktop (please complete the following information):**
 - macOS 11.6 Big Sur
 - Chrome
 - Version 94.0.4606.71 (Official Build) (x86_64)

extesy:
This seems to be a duplicate of #687. The settings are currently stored in localStorage which apparently gets cleaned when the cache is removed.

extesy:
https://github.com/extesy/hoverzoom/commit/6988beeb48ffde1d1f81a3efb1ef52eca5007cfd adds import/export functionality, will be available in the next release.

Issue #687, opened 2021-02-13
Option to export/import settings


ibLeDy:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I reinstall my OS or reinstall my browser, and I have to configure the settings manually each time.

**Describe the solution you'd like**
Buttons to be able to export and import all the settings.

merlin555:
Hi,

Please build a function to save the settings.
Thank you !

Make so on.
Very good addon!

Greetings..

Merlin

subsubsubsubsubsub:
The feature from 2021 has not been added

extesy:
https://github.com/extesy/hoverzoom/commit/6988beeb48ffde1d1f81a3efb1ef52eca5007cfd adds import/export functionality, will be available in the next release.

PR #1457, opened 2024-11-08
Add import/export function

LiliaDoe:
Adds an import and export box under the "Advanced" tab.
- Exports options in a json string format
- Has checks for valid HZ+ JSON string when importing
- Shows changed options upon importing (glows like when changing options)

Remedies #804 and #687

firefox_39Oah05hIH


extesy:
Very useful, thank you!

Issue #1318, opened 2024-03-07
Feature request: Allow "Enter" to be a key that locks the image on screen


dzek69:
**Is your feature request related to a problem? Please describe.**
I'm coming from other extension with similar functions (Imagus probably, I already forgot), it has the same lock image on screen feature, but it was under the Enter Key (either by default or I set it like that). I'm so used to do it even after half a year of using Hover Zoom + I still press it instead of the alternative I selected because of absence of Enter key in the drodown.

**Describe the solution you'd like**
Allow Enter key to be selected from the dropdown.

**Describe alternatives you've considered**
I tried `Shift` - it's a big key, next to Enter, but it's "not it".

**Additional context**
I think it wouldn't be a big deal to add Enter. Fingers crossed!

extesy:
Implemented by https://github.com/extesy/hoverzoom/commit/de95fae177778cfe10f7757c7aba991b53d9a70b

PR #1458, opened 2024-11-08
Add Enter as action key

LiliaDoe:
Requested by #1318

Initial tests show Enter now works as an action key. Not sure if there are cases it would cause issues.


extesy:
Well, it's not a default key, so whoever decides to assign Enter as their action key would do it at their own risk.

PR #1460, opened 2024-11-09
Add middle click and right click as action keys

LiliaDoe:
Fully adds right click and middle click as an action key. Before, right key was showing as an action key but was only being handled for 'lockImage'. This updates it so right click and middle click are completely useable as a key.

### Caveats and Issues I couldn't figure out:
- Middle click still does it's normal function, even after HZ+ function triggers. Right click doesn't
    - event.preventDefault() wasn't working to fix the issue, even after trying many things with it. 
- The hold delay is set to 150ms. It felt short enough to not wait, but it was long enough to not trigger accidentally
- Some buttons that don't require an image to be shown might trigger accidentally with normal use. E.g. toggle HZ+ when it's set to middle mouse click.
- The performance hit shouldn't be noticeable as a switch case is used to handle mouse events, but it is still triggered with every right and middle mouse click, even when they aren't selected for any action key.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -58,7 +58,7 @@ var factorySettings = {
     detailsLocation : 'none',
     fontSize : 11,
     fontOutline : false,
-    actionKey : -5,
+    actionKey : 0,
     toggleKey : 69,
     fullZoomKey : 90,
     copyImageKey: 67,
--- a/js/common.js
+++ b/js/common.js
@@ -58,7 +58,7 @@ var factorySettings = {
     detailsLocation : 'none',
     fontSize : 11,
     fontOutline : false,
-    actionKey : 0,
+    actionKey : -5,
     toggleKey : 69,
     fullZoomKey : 90,
     copyImageKey: 67,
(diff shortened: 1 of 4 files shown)


LiliaDoe:
Do you want me to move some of the duplicate functions in popup.js and options.js to common.js? Like keyCodeToString(), initActionKeys(), and loadKeys(sel)?

extesy:
@LiliaDoe Yeah, that would be nice. Also I'm not sure this is a good idea:

> Due to the reworking of the special codes for mouse buttons, action keys previously set to 'none' or 'right click' won't work or will show blank.

I've always tried to preserve backwards compatibility as much as possible. Whatever new functionality is added, nothing should change for existing users. Forcing them to re-select their already selected action keys doesn't seem user friendly.

LiliaDoe:
Gotcha. I can use an array to select which mouse button was pressed instead.

LiliaDoe:
That should be backwards compatible now

PR #1463, opened 2024-11-10
Fix: Github user images not zooming

LiliaDoe:
image sources with user-image urls on github weren't zooming. This fixes it.

### DIFF
--- a/plugins/github.js
+++ b/plugins/github.js
@@ -12,9 +12,9 @@ hoverZoomPlugins.push({
             res.push(img);
         });
 
-        $('img[src*="user-images"]:not(.hoverZoomMouseover)').addClass('hoverZoomMouseover').one('mouseover', function () {
-            const img = $(this);
-            const src = img.attr('src');
+        $('img[src*="user-images"]').one('mouseover', function () {
+            let img = $(this);
+            let src = img.attr('src');
 
             hoverZoom.prepareLink(img, src)
         });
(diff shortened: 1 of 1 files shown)


plugins/github.js
```diff
@@ -12,12 +12,13 @@ hoverZoomPlugins.push({
             res.push(img);
         });
 
-        $('img[src*="private-user-images"]').each(function () {
-            let img = $(this);
-            img.data('hoverZoomSrc', [img.attr('src')]);
-            res.push(img);
+        $('img[src*="user-images"]:not(.hoverZoomMouseover)').addClass('hoverZoomMouseover').one('mouseover', function () {
```
extesy, line 15: I wonder why hacks like `addClass('hoverZoomMouseover')` are necessary. Shouldn't this class be always set automatically by `hoverZoom.prepareLink`?

LiliaDoe: 'hoverZoomLink' is set by prepareLink(). Should it set hoverZoomLink instead or in addition?

extesy: I'm not sure I understand the purpose of adding `.hoverZoomMouseover` and using it in a selector. Since you are adding it _after_ selecting, that should guarantee that none of the existing elements would have it. And since this plugin is only executed once, there's no point of adding this class because there noone to check after the plugin is done.

LiliaDoe: Ohhh, I understand what you're saying now. I thought that part was necessary as it didn't seem to work without it, but after troubleshooting, it turns out it was something else. It is removed as you suggested

PR #1474, opened 2024-11-24
Add max width and height setting

LiliaDoe:
- Adds option to set max width and height of a zoomed image
- Picked up where Nymver left off, as it was nearly finished.
- Solves #1300


LiliaDoe:
I'm going to use a more updated branch as a base.

PR #1475, opened 2024-11-24
Add max width and height setting

LiliaDoe:
- Adds option to set max width and height of a zoomed image
- Picked up where Nymver left off, as the options were nearly finished.
- Solves #1300


LiliaDoe:
Tried to fix my git and things just got worse. Either way, it works, it just brings up secrets in naver.js again

extesy:
LGTM. I didn't see any secrets in naver.js though.

Issue #850, opened 2021-12-25
The image does not fill the entire available area


MatiasMovie:
**Describe the bug**
The image does not fill the entire available area

**To Reproduce**
1. Go to https://boards.4channel.org/a/
2. Hover on image

**Expected behavior**
The image fills the entire available area

**Screenshots**
![image](https://user-images.githubusercontent.com/24187109/147394943-224170ef-5bf9-4657-8bce-7ca382a64cb6.png)

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: Chrome
 - Version: 96

PR #1479, opened 2024-11-27
Add padding options

LiliaDoe:
- Adds the option to specify custom padding, allowing images to use nearly all of the viewer's window if set to 0
- further fixes issue #850
- I learned why extra padding is added to the top of the viewer. It centers the image vertically when the image fills the screen. It has been readded
    - Not dividing padding in half centered the image best, I found.
- Also changes 'pictures' from overlay options from my previous PR to 'images' to match the language other options uses

```js
547          if (position.top < wndScrollTop + padding) {
548              position.top = wndScrollTop + padding;
549          }
```


extesy:
Looks good, thank you!

PR #1483, opened 2024-11-29
Fix slickdeals plugin

khenriks:
This updates it to work with changes to the site. It's also backwards compatible, and won't affect places the plugin is currently working.


extesy:
Thank you for the contribution, @khenriks!

PR #1489, opened 2024-12-05
Add mouse click action key for right/middle mouse button

### LINKED ISSUE #457: REQUEST: Download image with a mouse button or gesture
Would it be possible to add a mouse gesture or shortcut (e.g. pressing the mousewheel) to download the zoomed image instead of "S"?

EDIT: Tried several remapping solutions but failed. I wanted to use the middle mouse button but not lose the normal functionality of it (closing tabs etc.). Ultimately I ended up using a program "X-Mouse Button Control", with a function called "Button Chording", where I mapped the RMB + MWheelDown combination to "s" (which is my download button in Hoverzoom). Not ideal, but kind of works. Maybe I should get a mouse with more buttons.

### LINKED ISSUE #1163: Activate Hover Zoom+ with Right Click not working in Edge
First option under the Action Keys. If I set Activate to Right Click, it doesn't seem to do anything. Toggling on/off does nothing. As soon as I change this to Ctrl, Shift, or any other key, click Save, that key works without even having to reload the page. 

Tested on two machines with Edge (114.0.1823.51 and 114.0.1823.55), both on Windows 11
The second machine only has two other extensions and they are both disabled.

### LINKED ISSUE #1190: Option to have middle mouse click as action key
I'm using mouse with one hand, to zoom in images using "Lock image on screen" feature you need both hands - for press some other key on keyboard without macros. 

That's very unintuitive and annoying.

##Would love option to have middle mouse click as action key.

### LINKED ISSUE #1487: [regression] Action key > Lock Image on click > right click not as reliable
**Describe the bug**
in recent builds, when the settings `advanced > Action key > Lock Image on click` is set to `right click` :
sometimes the right click menu will appear preventing the lock image behavior.
at other times it will work, but then once the image is locked, the right click doesn't trigger the contextual menu anymore.

**To Reproduce**
set ``advanced > Action key > Lock Image on click > right click`, then hover images, right click. sometimes it will fail. if it works, try to enable the right click menu : it doesn't appear
 
**Expected behavior**
in previous builds (I believe before commit #1472), it was working flawlessly : 
. RMB would lock the image without the contextual menu appearing
. once image is locked, a new RMB would trigger the context menu so we can interact with the locked image

`right click` to lock image is the default in Imagus, and it's useful for keyboardless browsing. Access to the right click menu is important to use addons like web clippers, image annotation, etc...

**Desktop (please complete the following information):**
 - win11 23h2
 - chromium 131
 - H+ Version 1.0.223

LiliaDoe:
- Adds right/middle mouse button click as an action key, rather than a hold
- Both a hold and a click can be used, on the same mouse button, to activate separate actions
- Fixes some return statements not returning the same values as before PR #1460
- Right click to lock image definitely works now. 
- Having middle and right click actions also works. E.g. activate on middle click hold and holding right click to zoom works as intended
- Added an option for users to define how long they need to hold before hold action activates. Also defines how quick they have to click the button. E.g. 250ms hold time means they have to press and release the button within 250ms for it to be a click and not a hold action
- Fixes  #1487
- Resolves #1190
- Resolves #457
- Might fix #1163

I've tested a majority of the action keys with right click and middle click. Lock works as it did before. If using both click and hold on right click, it's possible to still pull up the context menu on a zoomed image, depending on which action keys are set to right click.


LiliaDoe:
Everything looks correct, and what I have tested runs as expected. Performance seems fine from what I've tested. If any bugs or performance issues are found though, let me know and I'll work on fixing them~

js/hoverzoom.js
```diff
@@ -1083,43 +1083,84 @@ var hoverZoom = {
             }
         }
 
-        let longRightPressTimer; // create timer
-        let longMiddlePressTimer; // creates separate timer so they don't interfere
-        let longRightPress = false;
+        // create right and middle button timers to handle them separately
+        let longRightPressTimer;
+        let longMiddlePressTimer; 
+        // TODO: Prevent middle click from triggering when it's an action key
```
extesy, line 1089: Is this still TODO or is this comment a leftover from the early development?

LiliaDoe: It's something that is todo, but I couldn't figure it out yet

LiliaDoe:
I went through and tested what I could think of. More testing is always appreciated, but what I've tested and checked work as expected now.
- I had redo how it used preventDefault() so right and middle click would affect each other. 
- Right click to lock image definitely works now. 
- Middle click preventDefault() works too
- Fixed options not saving initially, requiring the button be pressed twice

extesy:
Since I just pushed a release and won't be doing new releases until 2025, I'm going to just merge this PR as-is and let it bake for 2-3 weeks. It will give us enough time for testing, and if you discover any new issues in the meantime, feel free to create follow-up PRs with fixes.

extesy:
@LiliaDoe There is a bug introduced in this PR. Hoverzoom shows the image even if the cursor moved past the link while the image was being prepared. This is extremely annoying because if for example I move the cursor across the page and it accidentally crosses some image link, it will be zoomed even if I don't keep the cursor on top of that link. Can you please take a look? Otherwise I'll have to revert this PR until the problem is resolved.

LiliaDoe:
@extesy I'll take a look at it. Are you using any mouse buttons as action keys?

extesy:
No keys, purely a mouse movement.

LiliaDoe:
I'm not sure what would be causing it in this PR. What you're describing happens to me once in a while, but it's very inconsistent. Does the image disappear again when you click? I have a PR that reverts most of the action key changes from this PR which may be the cause of the bug.

PR #1505, opened 2024-12-15
New action key: ban an image, video or audio track

GrosPoulet:
A new action key (default value: "B") can be used to prevent an image, video or audio track from being zoomed (and played) next time it is hovered. 

image

Ban is not lifted by just restarting browser. So in Advanced options, a new button is available in order to reset banned items so they can be zoomed again.

image

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -747,6 +747,18 @@
         "message": "On save, replace original filename by:",
         "description": "[options] Replace original filename when downloading option"
     },
+    "optResetAllBannedImagesTooltip": {
+    	"message": "After reset, all banned images, videos & audio tracks can be zoomed again",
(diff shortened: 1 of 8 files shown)


extesy:
@GrosPoulet This is a curious feature. What is the use case? Why would you want to ban a single image but not an entire website?

GrosPoulet:
@extesy This curious feature will be useful when i push improvments that will result in more images being zoomed, including background images. And sometimes, depending on website design, a background image "gets in the way", so user can ban it, et voilà!

extesy:
Be careful with that. Most users won't know about the "ban a single image" feature and when background images suddenly begin to unexpectedly pop up everywhere, it will be very annoying to many users.

PR #1515, opened 2024-12-29
Fix: images showing when mouse is off image link

LiliaDoe:
- This might fix the issue described in https://github.com/extesy/hoverzoom/pull/1489#issuecomment-2564391188

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1156,8 +1156,18 @@ var hoverZoom = {
                     lockImageKey(event);
                     return false;
                 case options.toggleKey: {
-                    let returnStatement = toggleKey() ? true : false;
-                    return returnStatement;
+                    options.extensionEnabled = !options.extensionEnabled;
+                    if (!options.extensionEnabled) {
+                        // close zoomed image or video
+                        viewerLocked = false;
+                        if (hz.hzViewer) {
+                            stopMedias();
+                            hz.hzViewer.hide();
+                        }
+                        if (imgFullSize) {
+                            return false;
+                        }
+                    }
                 }
                 case options.fullZoomKey:
                     if (!fullZoomKeyDown) {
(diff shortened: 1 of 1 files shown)


extesy:
This PR fixed the problem, so I'm merging it.

LiliaDoe:
Glad to hear!

extesy:
@LiliaDoe I spoke too soon, I'm still seeing this problem. Repro steps (can repro on reddit as well but this one is easier):
1. Open https://forums.crateentertainment.com/t/announcing-fangs-of-asterkarn/129879
2. Set "delay before displaying a picture" to some reasonably high value.
3. Briefly move the cursor into the picture and immediately move it far away.

Observed: the zoomed mage is still displayed even though the mouse is no longer hovering over it.
Desired: only display the image if the mouse is hovering over it.

LiliaDoe:
Ok, I can reproduce it consistently now. For the sake of the bug, it seems like a flag or variable is set to true once an image is zoomed the first time and not changing to false once the image stops being zoomed. My tests show (for me):
- Clicking on the page resets the "flag"
- Once an image on the page is zoomed, the "flag" is turned on for whatever triggers the bug
- When the "flag" is on, hovering around half the delay time and moving the mouse off the image will cause it to zoom

LiliaDoe:
I believe I found the source of the issue:
https://github.com/extesy/hoverzoom/blob/4cb6b61a03bdbee5a86eb8622db98368cf64b399/js/hoverzoom.js#L1074-L1086

changing it back to how it was prior made it behave as expected for me:
https://github.com/extesy/hoverzoom/blob/6b9717edd0d8b5da453381bbe60470bd2ca3b2ea/js/hoverzoom.js#L1071-L1074

While it's not my code, I will work on a solution

extesy:
Thank you for finding the problem. I thought I bi-sected the problem to your commit, but I guess I made a mistake while doing that. Please accept my apologies!

PR #1516, opened 2024-12-31
Fix: loadFullSizeImageImageTimeout not clearing when mouse leaves image

LiliaDoe:
- This fixes what #1515 was trying to fix
- setLoadImage used to set loadHoveredImage within callback so timeout can be set outside of callback. This makes clearTimeout use the correct timeout ID
- There is probably a smarter way of fixing this, but this works

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1072,8 +1072,8 @@ var hoverZoom = {
                             const delay = actionKeyDown || explicitCall ? 0 : (isVideoLink(srcDetails.url) ? options.displayDelayVideo : options.displayDelay);
                             loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
                             
-                            /** Temporarily removing until a better fix is found
-                            if (audioSrc) {
+                             // Temporarily removing until a better fix is found
+                            /*if (audioSrc) {
                                 chrome.runtime.sendMessage({action:'isImageBanned', url:audioSrc}, function (result) {
                                     if (!result) {
                                         loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
@@ -1085,8 +1085,8 @@ var hoverZoom = {
                                         loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
                                     }
                                 });                               
-                            }
-                            */
+                            }*/
+
                             loading = true;
                         }
                     } else {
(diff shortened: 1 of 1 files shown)


js/hoverzoom.js
```diff
@@ -1067,23 +1072,28 @@ var hoverZoom = {
...
                                     if (!result) {
-                                        loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
+                                        setLoadImage(true);
                                     }
                                 });                               
                             }
+                            
+                            // if the action key has been pressed over an image, no delay is applied
+                            const delay = actionKeyDown || explicitCall ? 0 : (isVideoLink(srcDetails.url) ? options.displayDelayVideo : options.displayDelay);
+                            if (loadHoveredImage) loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
```
extesy, line 1096: I'm not quite sure I understand how this works. Isn't `chrome.runtime.sendMessage` asynchronous? I mean it returns immediately and at some later point invokes a call back function. Which means line 1096 could check `loadHoveredImage` value before the callback has opportunity to set it.

LiliaDoe: That is true, and that is also why the bug was happening, i think. The timeout id being set by setTimeout wasn't always being cleared by clearTimeout due to it being async. I'm sure there is a better way of solving this

extesy: I think it's better to temporarily remove that entire message sending block until a proper solution is found. Right now the value of `loadHoveredImage` would be reused from the _previous_ image (because the current image would only set it after it was already checked) which is not correct.

LiliaDoe: If it's temporary, I could comment it out for now like
```js
/**
old code
*/
```

extesy: I'd imagine @GrosPoulet would want to make it work at some point so commenting out seems fine to me.

GrosPoulet: @extesy @LiliaDoe Happy New Year :tada:
I uncommented L1076-1088 and i can not reproduce the issue on Reddit or any other website. I have a 1 second delay before picture display.
My code:

`                                                        
                            // Temporarily removing until a better fix is found: sendMessage is async so it can't be used to set local variables
                            
                            if (audioSrc) {
                                chrome.runtime.sendMessage({action:'isImageBanned', url:audioSrc}, function (result) {
                                    if (!result) {
                                        loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
                                    }
                                });
                            } else if (src) {
                                chrome.runtime.sendMessage({action:'isImageBanned', url:src}, function (result) {
                                    if (!result) {
                                        loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
                                    }
                                });                               
                            }

                            loading = true;`

extesy: @GrosPoulet I posted an easy repro steps in https://github.com/extesy/hoverzoom/pull/1515#issuecomment-2566011814 - please take a look. Anyhow, the code as you wrote it couldn't possibly work due to async nature of sendMessage() function.

PR #1517, opened 2025-01-05
Update for "ban image" functionality (#1516)

GrosPoulet:
- when an image (or video & audio track) is banned:
   - url of banned image is stored in background page local storage
   - a message is sent to all tabs with updated list of banned urls
- removal of async call to background page: check for banned image is now performed synchronously in each tab


GrosPoulet:
@extesy looks like i have lost the ability to merge my PR ???

js/background.js
```diff
@@ -429,27 +430,38 @@ function banImage(message) {
...
+    var request = {action:'bannedImagesListChanged', 'list':list};
+    chrome.windows.getAll(null, function (windows) {
+        for (var i = 0; i < windows.length; i++) {
+            chrome.tabs.query({windowId: windows[i].id}, function (tabs) {
+                for (var j = 0; j < tabs.length; j++) {
+                    chrome.tabs.sendMessage(tabs[j].id, request);
+                }
+            });
+        }
+    });
```
extesy, line 464: This block is duplicated in `sendOptions()` function. It would be best to rename that one to something like `broadcastMessage` instead of duplicating the code.

js/hoverzoom.js
```diff
@@ -1066,26 +1067,24 @@ var hoverZoom = {
...
-                            } else if (src) {
-                                chrome.runtime.sendMessage({action:'isImageBanned', url:src}, function (result) {
-                                    if (!result) {
-                                        loadFullSizeImageTimeout = setTimeout(loadFullSizeImage, delay);
-                                    }
-                                });                               
-                            }*/
+
+
+
```
extesy, line 1078: too many empty lines

extesy:
Yes, I wanted to have opportunity to review the code before it's merged instead of doing post-merge fixes.

extesy:
@GrosPoulet If you still want this PR to be merged, please resolve conflicts. I'm going to be making more changes for Manifest V3 compatibility, so if it's not merged soon then it will be progressively harder to resolve conflicts.

js/hoverzoom.js
```diff
@@ -2643,12 +2639,27 @@ var hoverZoom = {
...
+        // get list of banned image, video or audio track urls
+        function loadBannedImages() {
+            chrome.runtime.sendMessage({action:'sendBannedImages'});
+        }
+
+        // check if url of image, video or audio track belongs to ban list
+        function isImageBanned(url) {
+            if (!url) return false;
+            reutrn !!bannedImagesList[url];
+        }
```
extesy, line 2651: @GrosPoulet Is `bannedImagesList` an array (list) or a set? According to https://stackoverflow.com/questions/2430000/determine-if-string-is-in-list-in-javascript if it's a list then you should use `.include` to check membership, and if it's a set then you should use `.has`.

GrosPoulet: @extesy bannedImagesList is a set (key:value), urls are used as keys

extesy: key:value is a dictionary (aka a map), not a set. Are values used for anything?

GrosPoulet: value = location (page url) where banned image was found

PR #1522, opened 2025-01-19
Manifest V3 migration

extesy:
https://developer.chrome.com/docs/extensions/develop/migrate


extesy:
@LiliaDoe @GrosPoulet This branch is at the state that it currently loads as a valid Manifest V3 extension and actually works for a couple sites that I tried. Feel free to experiment with it, I need all the feedback I can get before I merge it into main branch.

LiliaDoe:
Will do! Would you prefer it tester on chrome? Else, I'll test it mainly on firefox

extesy:
Testing in Firefox would be perfect. I mostly use Chrome so it's good to have some coverage between different browsers.

LiliaDoe:
My testing thus far has found a couple things. 
- The first is an issue with the manifest for Firefox.  I made a PR with a fix that should allow Chrome and Firefox to work. 
- The second is a reoccuring error: " Error: Could not establish connection. Receiving end does not exist." It's most consistent on https://www.furaffinity.net/, but I found it occurs on https://www.ebay.com/ upon initial load where it then goes away after the page finishes loading

extesy:
Thank you for testing! I also noticed this "Receiving end does not exist" error when saving options but I haven't yet been able to figure out what is causing it.

extesy:
[Migration docs](https://developer.chrome.com/docs/extensions/migrating/to-service-workers) say that unlike background pages, service workers could shut down at any time and lose all their registered listeners. So maybe that is what's causing the problem.

LiliaDoe:
I was seeing the error on the master branch too in reddit, while debugging that Reddit video issue

extesy:
@LiliaDoe I've made some fixes recently - the options can be saved now without errors but they don't seem to be updating other pages immediately, so a reload is required. Other than that, everything seems to work. Can you please re-check? I think I can live with options requiring a reload if it lets us restore the extension in Chrome web store.

LiliaDoe:
@extesy https://www.furaffinity.net/ still does not work, showing the same error. The error occurs on many other sites, but without affecting functionality. Options can be saved for me as well, despite the connection error showing

LiliaDoe:
@extesy I see it working on Chrome too, but the error still persists on Firefox

LiliaDoe:
I found a separate issue within firefox. In the permissions, every url that has a plugin for it is listed 
image

extesy:
@LiliaDoe I fixed `importScripts` error in Firefox, please check it out.

LiliaDoe:
@extesy That fixed the listener error, so now the images actually zoom! A new error shows up now however:
```
Error: Promised response from onMessage listener went out of scope background.js:292:26
```

extesy:
> The worker still stops when inactive

That's what service workers do and it's causing most of the problems for this migration.

> A new error shows up now however

Does it cause any practical issues with functionality or it's purely a log noise?

LiliaDoe:
> That's what service workers do and it's causing most of the problems for this migration.
Ah, I read about that behavior just being a service worker thing. Becoming less fond of mv3 the more I've worked on it.
> Does it cause any practical issues with functionality or it's purely a log noise?
Log noise. I double checked today on FA, and the error is no longer showing, soooo maybe a restart of the browser fixed it? I'll consider it resolved either way.

LiliaDoe:
I tested it on a variety of art sites, including pixiv and deviant art, and they all appear to work as before, with no errors now. Everything looks good on my end!

extesy:
Sounds good. I'm going to merge this branch and publish a new build but only for Chrome because it got disabled there. After that I'll wait for some time in case there are any serious bugs found, and if not then I'll publish it for Firefox and Edge. Thank you for testing!

sinkingpoint:
@extesy I note the FAQ on the settings page still tells folks to "In Options -> Advanced, activate "Allow rewrite of headers for HTTP(S) request or response"", but this PR seems to have removed that option - is there a new method in the manifest v3 world?

extesy:
@sinkingpoint That FAQ entry is not applicable anymore and should be removed. In MV3, headers are always updated according to a static ruleset.

PR #1529, opened 2025-02-02
Fix right click not opening context menu when set to lock image 

### LINKED ISSUE #1487: [regression] Action key > Lock Image on click > right click not as reliable

LiliaDoe:
This fixes context menu not showing up if right click is set to "Lock Image" and the image is currently locked. It should function like it use to now.

- fixes #1487

Also, I'm not sure where to report this, but the Storage API migration broke saving and the options UI on Firefox. I'm working on a fix for it, but I am not able to figure out how to get storage.sync.set to work on firefox yet


extesy:
> I am not able to figure out how to get storage.sync.set to work on firefox yet

Hmm, that's weird, the API is supposed to be the same across browsers - https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/sync

extesy:
Is this approach compatible both with Firefox in Chrome? If yes, do you mind sending a PR to fix it?

LiliaDoe:
I'll experiment with it and open a PR with it if it does work, sure!

LiliaDoe:
So, it was not working because firefox needs an ID in the browser_specific_settings section of the manifest for storage.sync: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings

Everything else just worked after that

extesy:
Wow, interesting, thank you for researching this. I was hoping that all browsers are more or less the same these days but I guess there are still some differences.

LiliaDoe:
Happy to help!

PR #1546, opened 2025-02-22
Add migrate old settings button

### LINKED ISSUE #1541: Recent update has reset plugins and sites settings
Checked on two separate PCs and my plugins and sites settings have been reset on both instances after todays update(s).

LiliaDoe:
- Should be a fix for those who's settings didn't migrate
- Working on a fix to migrate options function too

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -772,12 +772,16 @@
         "description": "[options] Enable debug option"
     },
     "optMigrateOldSettingsTooltip": {
-        "message": "Manually migrate settings from old save location",
+        "message": "Manually migrate settings from old save location. A save is needed after.",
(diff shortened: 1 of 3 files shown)


LiliaDoe:
localStorage returns undefined, and I'm not sure if that's fixable. I'm closing this until this I figure that out

LiliaDoe:
I figured out the storage system for a temporary addon was different than a signed addon in firefox, which caused my error. After fixing that part, the manual migrate button worked as expected. 

That said, if you could look through the option for it to make sure it looks how you'd like it, that'd be great! I added a warning to the button, since pressing it will overwrite the user's current settings:
image

extesy:
Looks great, thank you!

PR #1561, opened 2025-03-02
Make lock image not start zoomed in

### LINKED ISSUE #1374: Option for locked images/videos to maintain size (fit screen) and mousewheel to continue to control video position
**Is your feature request related to a problem? Please describe.**

Often when I am changing the video position with the mouse wheel, while hovering over a link, if the mouse moves even a little, it reloads the video because HoverZoom+ seem to think it is a new link to load, causing the video to restart (video linked below shows this behavior). This happens even when the mouse cursor maintains position over the link text.

This appears to happen most often with twitch video clips in Firefox at https://old.reddit.com/r/LivestreamFails/
Doesn't appear to happen with v.redd.it videos.

On Chrome, it seems to happen less often but it is inconsistent (happens for some twitch clip links and not for others - see video below).

Locking the video with action key "**Lock image on screen**" helps prevent this, but when you "Lock image on screen" it makes the image/video go to full size, and reverts the mousewheel to controlling the size of the image/video.

**Describe the solution you'd like**
Allow the option to have locked video/image just fit screen (same size they are when you hover), and still allow mousewheel to control video position. Just having it maintain size would fine as well but still allowing mouse wheel to control position would be nice.

**Additional context**
I wasn't sure if to make this a bug or feature request, I suppose it is bit of both.

Below is a video of the behavior in Google Chrome - you can see it working fine with one link, and doing the reloading of video with any movement in the second link.

https://github.com/extesy/hoverzoom/assets/243932/841b1af8-95cf-4da1-9b3d-aea5dd53bb28

LiliaDoe:
- Should resolve #1374
- Changes "Lock Image" so the first press will make the image fit the screen and the second press will make the image larger than the screen, on images large enough for it.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2810,17 +2810,15 @@ var hoverZoom = {
                 zoomFactorFit = width / srcDetails.naturalWidth;
                 lockViewer();
             }
-            else {
-                if (zoomFactor > 1.1 * zoomFactorFit || zoomFactor < 0.9 * zoomFactorFit) {
-                    // restore zoom factor such as img or video fits screen size
-                    zoomFactor = zoomFactorFit || parseInt(options.zoomFactor);
-                } else {
-                    // zoom factor = default
-                    zoomFactor = parseInt(options.zoomFactor);
-                }
-                posViewer();
-                panLockedViewer(event);
+            if (zoomFactor > 1.1 * zoomFactorFit || zoomFactor < 0.9 * zoomFactorFit) {
+                // restore zoom factor such as img or video fits screen size
+                zoomFactor = zoomFactorFit || parseInt(options.zoomFactor);
+            } else {
+                // zoom factor = default
+                zoomFactor = parseInt(options.zoomFactor);
             }
+            posViewer();
+            panLockedViewer(event);
         }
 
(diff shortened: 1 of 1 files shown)


LiliaDoe:
"potential simplification"? Does the bot display this suggestion somewhere?

LiliaDoe:
"potential simplification"? Does the bot display this suggestion somewhere? @gemini-code-assist

LiliaDoe:
> That's a great question! Yes, the bot (SonarQube in this case) usually displays its suggestions, including "potential simplification" hints, directly in the SonarQube Cloud interface. If you follow the link provided by the bot in its comment, you should be able to see the detailed analysis and specific code locations where it believes simplification is possible. Let me know if you have trouble finding it!

@gemini-code-assist There are no suggestions in the SonarQube Cloud interface. I am unable to find these simplifications

LiliaDoe:
@gemini-code-assist Show or tell me the potential simplification.

LiliaDoe:
That feels like it would be a very heavy line of code. Sure it does the same thing in a single line, but idk if it is an improvement

extesy:
This was a fascinating conversation! I have no idea what the bot meant by that "potential simplification" 🤷🏻‍♂️

Issue #1578, opened 2025-03-21
[Bug]: Firefox 136.0.2 and Hoverzoom 1.1.0 now doesn't show icon in address bar.


MarkRH:
### Description

After updating to Hoverzoom+ 1.1.0 in my Firefox 136.0.2 I noticed the icon that shows in the address bar when there is something to zoom no longer shows.  I double-checked the settings and the option is checked.  Tried unchecking and rechecking it.

![Image](https://github.com/user-attachments/assets/d755371f-141e-446b-82bf-94ec30b88357)

Here's an example that shows an image being zoomed, but icon is missing in address bar.

![Image](https://github.com/user-attachments/assets/db3f5126-9c9a-40d3-978d-aa3331ac6d23)

If any additional info you need, let me know.

-Mark

### Location

https://www.markheadrick.com/

### Reproduction steps

In this example, visit https://www.markheadrick.com/ but it seems any website that has images to zoom behave the same way. Same thing here: https://www.reddit.com/r/firefox/

### Expected behavior

Show the icon in the address bar if it's supposed to.
### Screenshots

See above

### Browser

Firefox 136.0.2

### Extension version

1.1.0

MarkRH:
Decided to test in my Clean test profile with the same results.

LiliaDoe:
@extesy This is a bug (I think) from MV3 migration that was missed.
This gets called as expected still:
 ```
chrome.action.setIcon({tabId: tab.id, path: '../images/icon19.png'});
``` 
The functionality of the action icon is now in the extension toolbar dropdown itself it seems. I didn't notice an icon on chrome either. A user can pin HZ+ to the toolbar as a workaround for now.

extesy:
According to https://developer.chrome.com/docs/extensions/reference/api/action#icon setting an icon from service worker requires using [offscreen canvas](https://developer.mozilla.org/docs/Web/API/OffscreenCanvas) API. I can take a look but no guarantees as to when.

LiliaDoe:
@extesy [offscreen canvas](https://developer.mozilla.org/docs/Web/API/OffscreenCanvas) is for dynamically generated icons in a service worker as opposed to [HTML canvas element](https://developer.mozilla.org/docs/Web/API/HTMLCanvasElement). The wording is a little confusing

LiliaDoe:
[...]

PR #1583, opened 2025-03-22
Change toolbar icon when on zoomable sites

LiliaDoe:
The deprecated code once used to put the icon in the address bar can now be used to change the icon of the toolbar icon instead. This PR reworks the code to add that functionality. Idea came from https://github.com/extesy/hoverzoom/issues/1578#issuecomment-2744565717:
- on sites with zoomable images, icon turns a purplish blue. Is similar to how the icon use to show in the address bar when on zoomable web pages
- updated option's message text to reflect new function

I wasn't sure what to make the new icon, so if you have a better idea for it, I'd be happy to change it


extesy:
Looks good to me, thank you!

PR #1590, opened 2025-03-23
Add use clipboard text when saving option

### LINKED ISSUE #1588: [Feature]: Save image as [content of clipboard]
### Description

I wonder if it is or would be possible to do this: on saving an image, replace the filename to the text currently in clipboard.

### Additional context

_No response_

LiliaDoe:
- Resolves #1588
- Very niche option, but figured it was simple enough to add
- Only making a PR in case someone else might want it
- If it doesn't fit HZ+ in general, I'd understand and will close it


extesy:
As far as I understand #1588, the request was for the opposite - use clipboard to determine a file name to save.

LiliaDoe:
Oh, that was the request. I misread completely

LiliaDoe:
Didn't mean to do that

PR #1594, opened 2025-03-24
Fix auto lock

### LINKED ISSUE #1586: [Bug]: Hoverzoom is no longer showing images full size when locking the image on the screen
### Description

Hello I'm having an issue with the latest update. Hoverzoom is no longer showing images full size when locking the image on the screen, I have to right click on the image and open in a new tab to see the full size. Using Chrome. Any fixes for this? Thanks for any info.

### Location

Every website

### Reproduction steps

Put the mouse over the image and try to make it full size no longer works with Lock the image on the screen

### Expected behavior

_No response_

### Screenshots

_No response_

### Browser

Chrome

### Extension version

Version 1.1.0

LiliaDoe:
This fixes auto lock so it behaves as if one pressed the "Lock Image" key. 

With auto lock on now, the image:
- fits to screen, unless "Locked images start zoomed in" is enabled
- will toggle between zoom states with additional presses of the "Lock Image" key
- uses the zoom factor for zoom amount unless "Use zoom factor for lock image" is disabled

I also fixed "Locked images start zoomed in" not working in general. 

Likely resolves #1586 (Not entirely sure the issue in that, but that's where I discovered the bugs fixed in this PR)


js/hoverzoom.js
```diff
@@ -2094,8 +2090,29 @@ var hoverZoom = {
                 hz.hzViewer.hide().fadeTo(options.fadeDuration, options.picturesOpacity);
             }
 
-            // The image size is not yet available in the onload so I have to delay the positioning
-            setTimeout(posViewer, options.showWhileLoading ? 0 : 10);
+            // The image size is not yet available in the onload so I have to delay the positioning 
+            setTimeout(() => {
+                posViewer();
```
extesy, line 2095: Why do you need to call `posViewer()` twice here? Should it be called just once at the end of this nested function?

LiliaDoe: The initial posViewer is needed so the width fits the screen size and zoomRatioFit gets calculated correctly. The second posViewer is to fix a visual issue where the image loads and then jumps to its locked position. It didn't seem to work correctly without both

LiliaDoe: I found a workaround where the image gets centered right away. That removes the need for the second ```posViewer()```

LiliaDoe: Making the images centered before the first ```posViewer()``` is called fixes the jumping issue, but the zooming issue persists

PR #1599, opened 2025-03-27
Make downloads permission optional

### LINKED ISSUE #1584: new 'manage downloads' permission
### Description

Just wondering what this is for. 

### Additional context

_No response_

LiliaDoe:
- Adds a new checkbox in the advanced tab of the options page which is used to request 'downloads' permission
- Both Firefox and Chrome get a prompt requesting the user to grant permission when the checkbox is ticked
- No longer requires 'downloads' permission to work
- Resolves #1584

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -575,6 +575,10 @@
         "message": "Add viewed pictures to the browser's history",
         "description": "[options] Add viewed pictures to the browser's history option"
     },
+    "optAllowMediaSaving": {
+        "message": "Allow saving media with action key",
+        "description": "[options] Allow saving media with action key option"
+    },
     "optFilterNSFW": {
         "message": "Exclude NSFW images (Reddit only)",
         "description": "[options] Exclude NSFW images (Reddit only) option"
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -576,8 +576,8 @@
         "description": "[options] Add viewed pictures to the browser's history option"
     },
     "optAllowMediaSaving": {
-        "message": "Allow saving media with action key",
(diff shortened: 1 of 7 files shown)


extesy:
What does the user see when they try to save the image and it won't save?

LiliaDoe:
Well, nothing without debug mode enabled. That's not a great experience now that you mention it. I made it an option because permissions.request has to be called from a user action, like a prompt or clicked button. Supposedly it's also possible from a keybind as long as its not an asynchronous keybind function. Maybe that would be better?

LiliaDoe:
I moved the options check before the saving functions are called to make it a little more efficient. Also, if the option is not enabled, an alert pops up, telling the user it's not enabled and what to do. I wasn't entirely sure how to properly word the alert, but it should work

extesy:
@LiliaDoe This need improvements to the initialization. The very first time I tried to save an image, it showed me the alert. That is despite me already having a permission, so when I opened the options page, the option was already checked. This will be very confusing to users who are asked to enable already enabled options. Basically, the default value should come from permission's current state, not just be always false.

LiliaDoe:
@extesy That makes sense. I'm not entirely sure if it's possible to check for permissions without going to the options page first. It was checked when you first went to the option page because it detected the permission was there, but I see how that is a confusing functionality. I can make it so it defaults to false and remains unchecked until a user goes in and checks the options themself. That way they would see the option was unchecked and wouldn't be confused

extesy:
If you only check permissions in the background script, how would you know when to show an alert from the main script?

LiliaDoe:
It would send a response. I made a PR with the idea #1600

PR #1600, opened 2025-03-28
Add permissions checking function to background

LiliaDoe:
This allows hoverzoom.js to check for 'downloads' permission before saving an image, in case the permission was already active or changed outside the options page.  This also allows an alert to be sent from the main script. The permission could instead be checked in hoverzoom.js ```init()``` so it's not called every time a user tries to save an image

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -3682,7 +3682,12 @@ var hoverZoom = {
         }
 
         function saveImage() {
-            cLog('check for media saving option');
+            cLog('check for permission');
+            chrome.runtime.sendMessage({action:'getPermissionsContains', permissions: ['downloads']}, 
+                function(response) {
+                    options.allowMediaSaving = response;
+            });
+                
             if (options.allowMediaSaving) {
                 saveImg();
                 saveVideo();
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -3685,18 +3685,16 @@ var hoverZoom = {
             cLog('check for permission');
             chrome.runtime.sendMessage({action:'getPermissionsContains', permissions: ['downloads']}, 
                 function(response) {
-                    options.allowMediaSaving = response;
-            });
-                
-            if (options.allowMediaSaving) {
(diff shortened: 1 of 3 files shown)


js/hoverzoom.js
```diff
@@ -3682,7 +3682,12 @@ var hoverZoom = {
...
 
         function saveImage() {
-            cLog('check for media saving option');
+            cLog('check for permission');
+            chrome.runtime.sendMessage({action:'getPermissionsContains', permissions: ['downloads']}, 
+                function(response) {
+                    options.allowMediaSaving = response;
+            });
+                
             if (options.allowMediaSaving) {
```
extesy, line 3691: The problem is that this is asynchronous call. The nested response handler will be executed _after_ the rest of `saveImage()` function completes. You need to move entire body of the function inside the response handler.

js/options.js
```diff
@@ -306,7 +306,6 @@ async function restoreOptions(optionsFromFactorySettings) {
     $('#chkShowDetailDimensions').trigger(options.showDetailDimensions ? 'gumby.check' : 'gumby.uncheck');
 
     $('#chkAddToHistory').trigger(options.addToHistory ? 'gumby.check' : 'gumby.uncheck');
-    $('#chkAllowMediaSaving').trigger(options.allowMediaSaving ? 'gumby.check' : 'gumby.uncheck');
```
extesy, line 309: Why remove it? The checkbox should be set according to the current permissions, not always off.

LiliaDoe: It handles this in the init, but it could be handled in restoreOptions() instead
https://github.com/extesy/hoverzoom/blob/6449651b0b84dfc90ddfe5e461ed06ca12f582f7/js/options.js#L488-L499
https://github.com/extesy/hoverzoom/blob/6449651b0b84dfc90ddfe5e461ed06ca12f582f7/js/options.js#L165-L169

extesy:
Thank you, this should work much better now for users who already have this permission.

PR #1608, opened 2025-04-11
Fix copy image and copy image url

### LINKED ISSUE #1607: [Feature]: Firefox browser lacks quick image copying operations compared to Google browser.
### Description

Image

Image

### Additional context

Firefox browser lacks quick image copying operations compared to Google browser.

LiliaDoe:
This attempts to fix "Copy Image" and "Copy Image URL" so it works in Chrome and Firefox. The main issue is Chrome requires "Access-Control-Allow-Origin" headers to be set. This is infeasible as it would have to be set for every site, unless it is set when the "Copy Image" key is pressed
- will fix #1607 once cors issue is resolved

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -3718,6 +3718,7 @@ var hoverZoom = {
                         func(blob);
                     },'image/png')
                 }
+                // Cross origin required to be set in order to convert canvas to blob
                 img.setAttribute('crossorigin', 'anonymous')
                 img.src = path;
             }
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -1212,12 +1212,7 @@ var hoverZoom = {
                     return;
                 }
                 case options.copyImageKey:
-                    if (isChromiumBased) {
-                        if (keyCode === options.copyImageKey) {
-                            copyImage();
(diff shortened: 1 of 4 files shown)


extesy:
@LiliaDoe I'm not familiar with this approach (setting CORS on `Image`). Have you confirmed that it works both in Firefox and Chrome? Also what does "begins to fix" mean?

LiliaDoe:
@extesy i meant it more like not fully fixing. On chrome,   it keeps throwing cors issues on sites without cors related headers e.g. pixiv

LiliaDoe:
@extesy After doing research, this is the best solution I can find. All solutions either involve the canvas api or the fetch api, both which throw the same error on Chrome for me. This solution allows Firefox to copy image and the usability in Chrome to remain the same. Currently, unless it's just on my end, "Copy Image" will throw a CORS error on any site that doesn't have the proper headers, either on their end or ours, and this change, while not fixing the error for Chrome, will allow Firefox to copy the image. There are extensions in Chrome that fix it, but I'm not sure it's possible on our end.

The one solution, which is not ideal, is to add the sites we/people want to copy images from most frequently to ruleset.json.

Issue #1604, opened 2025-04-02
[Bug]: After updating to 1.1.1, I suddenly couldn't preview Bilibili videos automatically.


hanson-reas:
### Description

This problem didn't hanppened before.

### Location

https://search.bilibili.com/all?vt=96667173&keyword=44&order=click

### Reproduction steps

Open the link and hover the mouse over any video cover. Before that, a preview of the complete video will appear. Now, it will only keep spinning, but nothing will appear.

### Browser

Chrome

### Extension version

1.1.1

extesy:
@GrosPoulet any idea how to fix it? I don't really understand what bilibili plugin is doing.

GrosPoulet:
I'll try to fix it next week

hanson-reas:
> I'll try to fix it next week我会在下周尝试修复它

thank you. But it still doesn't work in 1.1.2😂

hanson-reas:
> I'll try to fix it next week我将在下周尝试修复它

 may I ask if they will be fixed with in the next versions

extesy:
@GrosPoulet any hope you could take a look?

hanson-reas:
> [@GrosPoulet](https:github.com/GrosPoulet) any hope you could take a look?any hope you could take a look?
Thank you very much It has been resolved Thank you for your hard work

PR #1612, opened 2025-04-27
Fix for #1604

GrosPoulet:
Added explicit support for MPEG-DASH

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -635,9 +635,9 @@ var hoverZoom = {
             return (contentOverflows && overflowShown) || (alwaysShowScroll);
         }
 
-        const videoExtensions = new Set(['3gpp', 'm4v', 'mkv', 'mp4', 'ogv', 'webm']);
+        const videoExtensions = new Set(['3gpp', 'm4v', 'mkv', 'mp4', 'ogv', 'webm', 'm4s']);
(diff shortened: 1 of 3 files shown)


js/hoverzoom.js
```diff
@@ -635,9 +635,9 @@ var hoverZoom = {
             return (contentOverflows && overflowShown) || (alwaysShowScroll);
         }
 
-        const videoExtensions = new Set(['3gpp', 'm4v', 'mkv', 'mp4', 'ogv', 'webm']);
+        const videoExtensions = new Set(['3gpp', 'm4v', 'mkv', 'mp4', 'ogv', 'webm', 'm4s']);
         const videoExtensionsWithGif = new Set(['3gpp', 'gif', 'gifv', 'm4v', 'mkv', 'mp4', 'ogv', 'webm']);
-        const audioExtensions = new Set(['flac', 'm4a', 'mp3', 'oga', 'ogg', 'opus', 'wav']);
+        const audioExtensions = new Set(['flac', 'm4a', 'mp3', 'oga', 'ogg', 'opus', 'wav', 'm4s']);
```
extesy, line 640: @GrosPoulet How can `m4s` be both video and audio at the same time? None of existing extensions are shared between video and audio. How is having the same extension going to work?

GrosPoulet: Bilibili (and other websites) streams content using little M4S chunks of data: some audio chunks + some video chunks.
video.js player handles both perfectly:

bilibili

Issue #1618, opened 2025-05-04
[Feature]: Optionally disable looping of the images in a multi-image album.


coredumperror:
### Description

I recently discovered HoverZoom+ when looking for a replacement for Imagus, and I'm so glad I did! It does basically everything Imagus did, but is actually still maintained! Thank you.

The one thing I miss from Imagus, tho, is that when I'd seen every image in a zoomed album, pressing "next" (or scrolling the mouse wheel) would do nothing. That made it clear that I'd reached the end of the album, and I could move on to the next image on the page. 

This was especially handy on Pixiv, which might show you dozens of single or mult-image album links on a page, so knowing when you'd exausted an album and could move on the next one was really handy.

Would it be possible to add a checkbox to the settings that disables "looping back" to the first image in an album when you're at the end, please?

extesy:
Fixed by #1620, will be available in the next release

PR #1620, opened 2025-05-08
Added an option to disable looping albums

coredumperror:
The new option, `galleriesLoopEnabled`, is true by default. If disabled, navigating past the beginning or end of an album will do nothing, rather than looping back to the other side of it.

I created this PR after I realized that I could just do https://github.com/extesy/hoverzoom/issues/1618 myself. :)

Unfortunately, I have no idea how to create all the translations for the option description, so that'll have to be done before this PR can be accepted. Is there some script I can run that will auto-translate the `en` locale message that I did write?

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -243,6 +243,14 @@
         "message": "Navigate album with the mouse wheel",
         "description": "[options] Tooltip for use mousewheel to navigate albums option"
     },
+    "optGalleriesLoopEnabled": {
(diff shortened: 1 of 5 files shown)


extesy:
Thank you for the contribution, @coredumperror. Translations will be handled automatically via [crowdin](https://crowdin.com/project/hoverzoom) integration.

coredumperror:
Oh, cool! Glad I could contribute!

PR #1622, opened 2025-05-09
Pixiv plugin fix for PNGs, also improved performance

coredumperror:
This is a two-fold improvement for the Pixiv plugin:

1. This fixes a bug that caused PNG files to not be downloadable via the Save Image action. HZ would instead save the "master1200" thumbnail, which is a lower resolution JPG.

2. While fixing that, I also realized how to dramatically improve the performance of scrolling through a zoomed high-res gallery. Now, regardless of which format each file is in, they should load near-instantly while scrolling, especially with preloading enabled. 
 
Previously, PNG-based galleries would take about a half-second to load each image, during which the zoom would disappear, which was really jarring.

### DIFF
--- a/plugins/pixiv.js
+++ b/plugins/pixiv.js
@@ -63,7 +63,7 @@ hoverZoomPlugins.push({
                 return;
             }
 
-            cLog(`Fixing ${galleryUrls.length} imaages...`);
+            cLog(`Fixing ${galleryUrls.length} images...`);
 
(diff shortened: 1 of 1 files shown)


extesy:
Excellent, thank you!

coredumperror:
Glad tp help! Looking forward to a new release that has my two PRs applied. :)

PR #1636, opened 2025-06-12
Fix dimensions details

runer112:
The dimensions details should now display the image's natural size.

Apparently since its introduction by commit e891d02b7ab9080e522c5028a978df9d6987bb16, the dimensions details have divided the image's natural size by the the zoom factor. I'm not sure why this was done. *Multiplying* by the zoom factor would at least compute the displayed size (although I think this is a less desirable metric to show than the natural size). But dividing doesn't seem to produce any sensible metric.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2298,7 +2298,7 @@ var hoverZoom = {
         function getSrcDetails(link) {
             let details = {};
             if (!srcDetails.audio || srcDetails.video) {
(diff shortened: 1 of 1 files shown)


extesy:
Let's ask the original author. @GrosPoulet did you have any particular reason to implement your way of computing image dimension details?

GrosPoulet:
Probably a copy/paste mistake, zoom factor can be removed.

runer112:
Thank you for accepting this PR. I'm now using the updated extension distributed through the Chrome web store, and I'm pleased to confirm that the dimension details appear as expected!

Issue #336, opened 2018-01-27
bug? weird tab behavior since recently


baz357:
Something changed:

BEFORE:
Google image search is tab #4
I hover over an image
I press "t"
The new image is opened as tab #3
I press "t"
The image is closed, the Google image search in tab #4 is visible again

NOW:
Google image search is tab #4
I hover over an image
I press "t"
The new image is opened **as tab #1**
I press "t"
The image is closed, but **tab #1 is visible, not #4**

Is this a bug? I first noticed this unwanted behavior 10-14 days ago, and it's still the same after updating Chrome yesterday.

Current version: 1.0.125 (Win 10)

ivan2340:
This seems to be a bug, i also recently noticed this behaviour

extesy:
Something might have changed in Chrome again because tab opening logic hasn't been touched for a long time.

PR #1650, opened 2025-07-04
Always open new tabs next to the current tab

### LINKED ISSUE #1649: [Bug]: The "Open Image in a New Tab" action sometimes opens new tabs in subpar locations
### Description

Using the "Open Image in a New Tab" action causes tabs to appear:
- right before of the current tab (original intended behaviour)
- or in a (seemingly) random position anywhere before the current tab (the bug)

I feel that the new tab should always appear right *after* the current tab, with the `shift` key controlling whether the tab is made active or not. Since the action is called "Open Image in a *New* Tab", I feel the extension should not be overwriting the current tab index.

I think I have found what is causing this, I will open a pull request in a bit.

### Location

All websites

### Reproduction steps

- Open two browser windows, window `A` with 2 tabs and the other window `B` with 5 tabs
- Use the "Open Image in a New Tab" action on tab 5 in window `B`
- The new tab is created at at position 2 (instead of 5 or 6)

### Expected behavior

The new tab should always be opened in a tab next to the current tab.

### Screenshots

_No response_

### Browser

Firefox

### Extension version

 1.1.5

pranavmangal:
The [`chrome.tabs.query({active: true})`](https://developer.chrome.com/docs/extensions/reference/api/tabs#method-query) returns a list of active tabs across *all* browser windows. This means that picking `tabs[0]` doesn't necessarily correspond to the current window which can cause new tabs to be positioned in odd locations.

Additionally, the original implementation chose to open a new tab after the current tab only in case of background tabs (using `shift`). Non-`shift` uses opened the new tab at the current index, which had the effect of seemingly opening the new tab *before* the current tab. This has been simplified to always open a new tab *after* the current tab.

Fixes #1649 

Might also possibly be a fix for the odd behaviour explained in #336

### DIFF
--- a/js/background.js
+++ b/js/background.js
@@ -200,10 +200,8 @@ async function onMessage(message, sender, sendResponse) {
             });
             break;
         case 'openViewTab':
-            chrome.tabs.query({active: true}, function (tabs) {
-                message.createData.index = tabs[0].index;
-                if (!message.createData.active)
-                    message.createData.index++;
+            chrome.tabs.query({active: true, currentWindow: true}, function (tabs) {
+                message.createData.index = tabs[0].index + 1;
                 let url = message.createData.url;
                 if (url.indexOf('facebook.com/photo/download') !== -1) {
                     message.createData.url = 'data:text/html,<img src="' + url + '">';


extesy:
This looks very reasonable, thank you!

Issue #1645, opened 2025-06-27
[Website]: Yupoo


CobbPalahniuk:
### Website

https://yupoo.com/

### Explanation

Can you please add support for yupoo galleries? This is an example [https://ezfashion.x.yupoo.com/categories/390994](https://ezfashion.x.yupoo.com/categories/390994)

PR #1659, opened 2025-07-15
Plug-in for : Yupoo.com (#1645)

GrosPoulet:


GrosPoulet:
**Warning**: there is an issue with large collection of galleries: after scrolling, HZ+ viewer is not displayed in viewport. Instead, it is located too low on page resulting in truncated or hidden viewer.
**Workaround**: in Options : General, enable "center images" setting
@LiliaDoe since you worked on viewer management, do you have any idea on how we can fix this ?

extesy:
I'll wait for a few days for @LiliaDoe to respond before I merge this PR.

extesy:
Well... I guess we'll just have to fix it forward then.

PR #1724, opened 2025-12-22
Improvements for Pinterest plug-in

GrosPoulet:
Added support for videos

image


extesy:
@GrosPoulet Thanks for the update. I see that Gemini bot has a few good suggestions - can you take a look and apply where it makes sense?

GrosPoulet:
all Gemini suggestions applied and tested OK

Issue #845, opened 2021-12-19
Bug with album scrolling - goes through 1-2 photos per scroll notch


RoboChrist161:
**Describe the bug**
Album photos scroll too quickly. Can use arrow keys as temporary remedy. Saw this this issue had previously been noted in extesy#76 and fixed using an internal limiter, but that change was not committed to main branch:
https://github.com/Markussss/hoverzoom/commit/1996a6fc7b2af682d1244c29b31d0cb4ca07324b

**To Reproduce**
1. Hover over album
2. Scroll mouse down or up
3. Scroll through multiple photos, or back to the original photo in a 2-photo album.

**Expected behavior**
Expect to scroll to next photo with mouse wheel, rather than scrolling through 2+ photos.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version [Version 1.0.185]

**Additional context**
N/A

extesy:
I've noticed this sometimes but wasn't able to reliable reproduce the problem.

TheMikeyR:
@extesy 
I believe this is touchpad related behaviour. I'm using a mac and this issue exist on both external touchpad and the build-in. When using a logitech mouse the scroll acts normally. 
It is there always, and not something that occurs sporadic. 
I've included a video to demonstrate. 

Let me know if there is anything I can do to help out. 

https://github.com/user-attachments/assets/8bccf8e4-88f0-4cde-b339-04f8d61f2efc

extesy:
Thank you for the video, it looks like touchpad sends a lot of scrolling events since it is supposed to have a much higher scrolling accuracy compared to mouse. I'm not sure what to do about that and even if I try something, I won't be able to verify since I don't use a Mac.

Kellett:
I've had this issue for a long time. I think its something to do with my mouse (MX Master). The thumb scroll wheel (horizontal) works normally and will do a single photo per scroll. The normal wheel does loads. I tried fiddling with the firefox config settings for `mousewheel.default.delta_multiplier_ x/y/z` and `mousewheel.min_line_scroll_amount` but it didn't seem to help.

Once I swapped to FireFox via Flatpak the issue went away so I forgot about it. However I recently swapped to Wayland (Or it could be a new firefox version) and its back.

extesy:
Fixed by 6f192dcc489cf902dccbe6eb1675347a031f76f0

PR #1742, opened 2026-02-09
Fix for erratic scrolling

Kellett:
Attempted fix for https://github.com/extesy/hoverzoom/issues/845

When I had a better look into this issue it seems like FireFox turns my single scroll into multiple smaller ones that each fire documentOnMouseWheel. I added a cooldown to the function so it will only accept 1 scroll event per 350ms. This fixes the issue for me 99% of the time.
 
Maybe this should be put behind an advanced setting rather than "live" for everyone. I think the cooldown amount of 350 could probably be lowered as well if you think this is too slow.

### DIFF
--- a/js/hoverzoom.js
+++ b/js/hoverzoom.js
@@ -2786,7 +2786,7 @@ var hoverZoom = {
                 posViewer();
                 panLockedViewer(event);
             } else {
-                if (now - lastScrollTime < 350) { 
+                if (now - lastScrollTime < options.scrollWheelCooldown) { 
                     event.preventDefault();
                     return;
                 }
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -499,6 +499,18 @@
         "message": "milliseconds",
         "description": "[options] Unit name (pixel) for hold time amount"
     },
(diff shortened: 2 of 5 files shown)


js/hoverzoom.js
```diff
@@ -2782,10 +2785,16 @@ var hoverZoom = {
                 zoomFactor = Math.max(Math.min(zoomFactor, 10), stepInit);
                 posViewer();
                 panLockedViewer(event);
-            } else if (imgFullSize) {
+            } else {
+                if (now - lastScrollTime < 350) {
```
extesy: This needs to be a proper config with default value set to 0 to avoid changing existing behavior. With roughly 400k users of this extension, I'm pretty sure there are some that rely on "press and hold a key" to rapidly scroll and this would break that use case.

Kellett:
I've updated the PR to put this change behind a setting. I've placed it in the Delay section:
image

The default is set to 0.

I wasn't sure what to do about the different language options.

Let me know if you want anything reworded or changed

extesy:
Looks good to me, thank you for the contribution!

PR #1749, opened 2026-03-03
docs: update Chrome Web Store URLs to new format

theluckystrike:
## What
Updates 2 Chrome Web Store URLs in README.md from the old format to the new format.

## Why
The old format URLs are being deprecated. This update ensures the links work properly.

## Testing
- Verified URL structure is correct


theluckystrike:
Thanks for the review @gemini-code-assist! 

The  (encoded space) from the old URL has already been removed in this PR. Looking at the diff:

- Old URL: `https://chrome.google.com/webstore/detail/hover-zoom%20/pccckmaobkjjboncdfnnofkonhgpceea` (note the space after "hover-zoom")
- New URL: `https://chromewebstore.google.com/detail/hover-zoom/pccckmaobkjjboncdfnnofkonhgpceea` (space removed)

The new Chrome Web Store format doesn't use spaces in the URL path, so the space character and its encoding have been properly removed. The extension ID `pccckmaobkjjboncdfnnofkonhgpceea` remains the same, ensuring the link works correctly.

PR #1760, opened 2026-03-18
Adds support for Vinted

kikoosvk:
Add preview support for Vinted item images on listing, homepage, and item pages, including lazy-loaded feed items.

image

### DIFF
--- a/js/hoverZoomVintedFetch.js
+++ b/js/hoverZoomVintedFetch.js
@@ -1,0 +1,143 @@
+// Vinted page-world fetch/XHR bridge.
+// Captures listing API payloads and replays wardrobe requests from the page
+// so Vinted sees the same first-party cookies and request headers as the site.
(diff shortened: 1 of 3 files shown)


extesy:
Thank you for the pull request! This is a huge plugin, probably the largest in the entire repo. What makes Vinted so special that it requires so much code?

kikoosvk:
Most plugins can get the full image with a simple URL rewrite or one small request. Vinted is harder because the image/gallery data is spread across SSR data, runtime API responses, lazy-loaded homepage blocks, and sometimes a wardrobe fallback for feed items. It also loads content dynamically as you scroll, so the plugin has to work lazily instead of just processing the initial DOM once.

So the size is mostly a reflection of Vinted’s frontend architecture, not because the plugin is doing anything unusually fancy.

PR #1780, opened 2026-05-24
Update messages.json

liveasx:


extesy:
@liveasx Localization fixes must be done via https://crowdin.com/project/hoverzoom otherwise they will all be overwritten.

PR #1781, opened 2026-05-28
Bluesky: Load profile pictures in posts and popups

dubhater:
Currently a user's profile picture only loads when visiting the profile. The URL looks like this:  ``https://cdn.bsky.app/img/avatar/...``

Profile pictures in posts and the popups that appear when hovering over a user name look like this:
``https://cdn.bsky.app/img/avatar_thumbnail/...``

Extend the URL replacement to match these URLs too.


extesy:
LGTM, thank you for contribution

